PR #3, opened 2019-05-07
Update the regex for csrf

bitdivision:
Ian noticed some issues when running this against v3. 

I'm not entirely sure why this has changed, but the included regex seems to work.


bitdivision:
As suggested by @iwootten this is probably to do with the compression change.

PR #7, opened 2019-09-12
Performance testing setup

LJBabbage:
Update to work with current release and add JSON for individual schema


requests/test_checkbox.json
```diff
@@ -1,40 +0,0 @@
-{
```
ajmaddaford: I think we should keep this test file as it's a nice short example, that can also be used when testing changes to the benchmarking code.

PR #8, opened 2019-09-19
Allow dynamic urls in test json 

iwootten:
**Motivation**

List collectors generate dynamic urls during a survey and the benchmark tool only handles a hardcoded list of urls.

**Changes**

Introduce a new json attribute 'redirect_route', which captures arguments from a url and supports their later use throughout the test json. HAR file output can be captured, transformed to json and then dynamic params can manually replaced with placeholders. See the example census_household request file for how it's used.


runner_benchmark/taskset.py
```diff
@@ -33,34 +34,52 @@ def replay_requests(self):
...
             if request['method'] == 'GET':
-                response = self.get(request['url'])
+                response = self.get(request_url)
 
                 if response.status_code not in [200, 302]:
                     raise Exception(
-                        f"Got a ({response.status_code}) back when getting page: {request['url']}"
+                        f"Got a ({response.status_code}) back when getting page: {request_url}"
                     )
+                if response.status_code == 302:
```
JamesGardiner: This block is identical to lines 74 - 78. Could be a function

runner_benchmark/utils.py
```diff
@@ -0,0 +1,25 @@
+from urllib.parse import urlparse
+
+
+def has_params(string):
+    return '{' in string and '}' in string
```
JamesGardiner: This function isn't required

runner_benchmark/taskset.py
```diff
@@ -33,34 +34,52 @@ def replay_requests(self):
         user_wait_time_min = int(os.getenv('USER_WAIT_TIME_MIN_SECONDS', 0))
         user_wait_time_max = int(os.getenv('USER_WAIT_TIME_MAX_SECONDS', 0))
 
+        redirect_params = {}
+
         for request in self.requests:
+            request_url = (
```
JamesGardiner:
```suggestion
            request_url = request['url'].format_map(redirect_params)
```

iwootten:
> The only issue was the manual step of adding in 'redirect_route'. A 5 person household with 2 visitors needed 18 manual additions into the JSON. I don't think this should stop this card merging, but reckon we should raise another card to look at this (if its even possible)

May be possible. Would require the extractor picking up on the first redirect of a list collector and substituting placeholders throughout the json which is extracted. 

Given the ids the collector creates are repeated this doesn't seem a massive overhead for the purposes of benchmarking at the moment. We have already have schemas for household/individual so it may not be worth spending additional time on it though.

PR #12, opened 2019-11-07
Use Kubernetes Job Kind and Store benchmark output in GCS bucket

MebinAbraham:
1. Previously the Kubernetes deployment would restart a pod once locust kills the pods once the test length has passed. This PR makes use of the Kubernetes `Job` Kind to run the task until completion/failure, meaning the pods will not get restarted once the locust run has finished. (Locust test length is specified by the locust options)

2. Add a script to push the locust benchmark outputs to a GCS bucket.

### How to review:
1. Test using: https://github.com/ONSdigital/eq-terraform-load-generator/pull/2
2. Ensure the pods are killed once the specified test length has passed.


k8s/helm/values.yaml
```diff
@@ -2,19 +2,18 @@
...
+memory: "2048Mi"
 
 host: ''
-requestsJson: requests/census_household_gb_eng.json
+requestsJson: 'requests/census_household_gb_eng.json'
 userWaitTimeMinSeconds: 1
 userWaitTimeMaxSeconds: 2
-locustOptions: -f locustfile.py --no-web -c 800 -r 10 --loglevel=CRITICAL --run-time=2h --only-summary
-
+locustOptions: '--locustfile locustfile.py --no-web --clients 800 --hatch-rate 10 --loglevel=CRITICAL --run-time=2h --only-summary'
```
ajmaddaford: `--locustfile` is not necessary since we renamed the main locustfile.

k8s/helm/values.yaml
```diff
@@ -2,19 +2,18 @@
...
+gcsOutputBucket: ''
+outputDirectory: 'manual-run'
```
ajmaddaford: Why `manual-run`?

Also, we should make use of nesting for associated values. For example these could be:

```
output:
  bucket:
  directory:
```

README.md
```diff
@@ -147,6 +148,8 @@ If you want to vary the default parameters Locust uses on start, you can specify
   - defaults to 1
 - userWaitTimeMaxSeconds - The maximum delay between each user's GET requests
   - defaults to 2 
+- gcsOutputBucket - Name of the GCS bucket in which the output should be stored.
+  - defaults to <project-id>-benchmark-outputs
```
ajmaddaford: Where is this defaulted?

k8s/helm/values.yaml
MebinAbraham: Removed.

k8s/helm/values.yaml
MebinAbraham: Removed default output dir name and added nesting for associated values.

README.md
MebinAbraham: Removed. Was thinking about Concourse.

README.md
```diff
@@ -155,7 +158,7 @@ helm tiller run \
 -    runner-benchmark \
 -    k8s/helm \
 -    --set requestsJson=requests/census_individual_gb_eng.json \
--    --set locustOptions="-f locustfile.py -c 1000 -r 50 -L WARNING" \
+-    --set locustOptions="-f locustfile.py --clients 1000 --hatch-rate 50 -L WARNING" \
```
ajmaddaford: `-f locustfile.py` should be removed.

MebinAbraham: Done.

scripts/extract_distribution_summary.py
```diff
@@ -17,7 +17,7 @@
         get_values.append(percentile_99th)
     if 'POST /questionnaire' in line:
         post_values.append(percentile_99th)
-    if 'Total' in line:
+    if 'Aggregated' in line:
```
MebinAbraham: Changes since locustio have been updated.

store_benchmark_outputs.py
```diff
@@ -0,0 +1,69 @@
...
+                max_wait=max_wait,
+                requests_json=requests_json,
+                locust_options=locust_options,
+                runner_version=runner_version,
+                timestamp=int(datetime.utcnow().timestamp()),
+            )
+        except Exception as ex:
+            raise Exception(f"Error occurred during file upload to GCS - {ex}")
+    else:
+        raise Exception('GCS Bucket not specified by `GCS_OUTPUT_BUCKET`')
```
JamesGardiner: Raising a bare exception isn't good style. ValueError may be more appropriate. Also, the error message isn't clear - something like 'GCS_OUTPUT_BUCKET is None' may be clearer.

store_benchmark_outputs.py
```diff
@@ -0,0 +1,69 @@
...
+                host=host,
+                min_wait=min_wait,
+                max_wait=max_wait,
+                requests_json=requests_json,
+                locust_options=locust_options,
+                runner_version=runner_version,
+                timestamp=int(datetime.utcnow().timestamp()),
+            )
+        except Exception as ex:
+            raise Exception(f"Error occurred during file upload to GCS - {ex}")
```
JamesGardiner: Catching the root exception class isn't good style. There are a small number of exception that could be raised by the google library during this operation, such as [NotFound](https://github.com/googleapis/google-cloud-python/blob/fd93452ddbe12b564c02a797cc462ae4e6dc7eb5/storage/google/cloud/storage/client.py#L279). They should be explicitly caught.

PR #13, opened 2019-11-08
Add performance investigations proposal

ajmaddaford:
Some potential performance improvements could be achieved in runner. We need to define a structured approach to these performance investigations to ensure that we are improving performance. For each improvement we need to clearly document the proposed change and the output of any benchmark tests that are done. 

This pull request sets out a proposed approach for this.


doc/performance-investigations/0001-use-grpc-for-datastore-access/summary.md
```diff
@@ -0,0 +1,28 @@
+# Use GRPC for Datastore access
+
+Use the GRPC option for Datastore access.
+
+https://github.com/ONSdigital/eq-survey-runner/tree/use-grpc-for-datastore-access
```
MebinAbraham: These will probably end up getting deleted at some point in the future when they become stale but probably worth having a record as long as this doc exists.

ajmaddaford: Yeah, fair point. We could prefix the branches with `performance-investigation-` to highlight they shouldn't be removed.

MebinAbraham:
We should repeat the tests for reliability?

doc/performance-investigations/0001-use-grpc-for-datastore-access/summary.md
MebinAbraham: 👍 It's worth adding that in this doc.

doc/performance-investigations/extract_distribution_summary.py
```diff
@@ -0,0 +1,28 @@
...
+        continue
+
+    values = line.split(',')
+    percentile_99th = int(values[9])
+
+    if 'GET /questionnaire' in line:
+        get_values.append(percentile_99th)
+    if 'POST /questionnaire' in line:
+        post_values.append(percentile_99th)
+    if 'Total' in line:
```
MebinAbraham: Total is not valid?
```suggestion
    if 'Aggregated' in line:
```

ajmaddaford: I've just checked in a clean virtual env and my file has "Total". 

On the repeating of tests, what do you suggest? I'm aware that if we need to run lots of tests it will slow down progress.

doc/performance-investigations/README.md
```diff
@@ -0,0 +1,29 @@
...
+1. Capture metrics - `./run.sh requests/<requests_file> https://<runner_url> <duration> investigation`.
+1. Copy the generated CSVs into the investigation folder.
+1. Update `summary.md` with the results.
+
+## Extracting performance metrics from CSVs
+
+You can use the `extract_distribution_summary.py` script to extract high level metrics from the Locust distributions CSV:
+
+```
+python extract_summary.py < baseline_distribution.csv
```
LJBabbage: Do you mean extract_distribution_summary as I don't have an extract_summary? Also this command might want to include `<the location of your csv file`> in the command as the extract_distribution_summary and csv wouldn't live in the same place

doc/performance-investigations/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Creating a new investigation
+
+1. Create a branch in runner prefixed with `performance-investigation` and make the proposed changes.
+1. In this repo create a new folder named `0000-descriptive-name-for-investigation` incrementing `0000` from the last used number.
+1. Copy `TEMPLATE.md` into the directory as `summary.md`.
+1. Update `summary.md` as appropriate.
+1. Capture baseline metrics - `./run.sh requests/<requests_file> https://<runner_url> <duration> baseline`.
+1. Deploy the investigation branch.
+1. Capture metrics - `./run.sh requests/<requests_file> https://<runner_url> <duration> investigation`.
```
LJBabbage: <duration in minutes/hours> maybe?

doc/performance-investigations/README.md
```diff
@@ -0,0 +1,29 @@
+# Performance Investigations
+
+## Creating a new investigation
+
+1. Create a branch in runner prefixed with `performance-investigation` and make the proposed changes.
+1. In this repo create a new folder named `0000-descriptive-name-for-investigation` incrementing `0000` from the last used number.
+1. Copy `TEMPLATE.md` into the directory as `summary.md`.
+1. Update `summary.md` as appropriate.
+1. Capture baseline metrics - `./run.sh requests/<requests_file> https://<runner_url> <duration> baseline`.
```
LJBabbage: In the summary.md we are looking to record the user wait time minimum; user wait time maximum; clients and hatch rate. But it's not obvious what they are set to, or how we change them from this readme. A more explicit command would help possibly

doc/performance-investigations/README.md
ajmaddaford: Script moved to scripts directory and Readme updated.

doc/performance-investigations/extract_distribution_summary.py
MebinAbraham: So as discussed, the latest version of locust seems to use `Aggregated` so could be worth upgrading but will leave it up to you.

As for the repeating of the tests, I think it worthwhile to at least test the changes twice to potentially reduce the chance that it is an anomaly?. I guess it could be that the tests are run 2-3 times but we only check-in 1 of the results on the basis that they are similar.

doc/performance-investigations/README.md
ajmaddaford: Readme updated.

doc/performance-investigations/README.md
ajmaddaford: Readme expanded.

doc/performance-investigations/extract_distribution_summary.py
ajmaddaford: I've added a line about running the tests more than once. I've opened a separate pr for updating the Python modules as it didn't seem related to a pull request adding documentation.

PR #16, opened 2019-12-27
Add run-benchmark ci task

JamesGardiner:
This PR adds a new directory, `ci`, that contains tasks for concourse pipelines.

To test, run the `run-stress-test` job of the performance testing stress-test pipeline https://github.com/ONSdigital/eq-concourse-peformance-testing/blob/stress-test/stress-test.yaml


JamesGardiner:
[Performance stress testing runner](https://trello.com/c/1in6xoZg/3403-performance-stress-testing-runner)

ci/run-locust.yaml
```diff
@@ -0,0 +1,56 @@
+platform: linux
```
ajmaddaford: This file should be called `run-benchmark.yaml`.

JamesGardiner: It originally was, but the implementation is locust specific.

ajmaddaford: Only `locustOptions` is locust specific, and that should also be fixed in another pr.

ci/run-benchmark.yaml
```diff
@@ -0,0 +1,55 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: theorf/google-cloud-sdk-helm
+inputs:
+  - name: eq-survey-runner-benchmark
+params:
+  IMPORT_EXISTING_PROJECT: true
```
ajmaddaford: We should consider removing the `IMPORT_EXISTING_PROJECT` functionality given that it's the default in the pipelines.

ci/run-benchmark.yaml
```diff
@@ -0,0 +1,55 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: theorf/google-cloud-sdk-helm
+inputs:
+  - name: eq-survey-runner-benchmark
+params:
```
ajmaddaford: I think it would be a good idea to list the params that this task uses here. This is highlighted as a common pattern at https://concourse-ci.org/task-step.html: 
> Looking into the task.yml, a common pattern is to list the param in the task definition with no value. This indicates that you expect the pipeline to provide the value.

JamesGardiner: Ok this makes good sense to me - even if just to document what params are expected by the task so that it is easier to `fly execute`. One question - what is your opinion on having the params default to empty strings (such as `PROJECT_ID:` versus a default variable expansion such as `PROJECT_ID: ((google.project_id))`? If the former then anytime the param isn't passed to the task, it will become an empty string. I'm assuming here that variable expansion will work in that part of the task file.

ajmaddaford: I think empty is best for now while we work through externalising all the tasks.

JamesGardiner:
> Now that cpu and memory is no longer configurable it should be removed from the Helm chart.

Done

ci/run-benchmark.yaml
```diff
@@ -0,0 +1,59 @@
...
+        k8s/helm \
+        --set requestsJson="${REQUESTS_JSON}" \
+        --set locustOptions="${LOCUST_OPTIONS}" \
+        --set host=https://${RUNNER_FULLY_QUALIFIED_DOMAIN_NAME} \
+        --set container.image=${DOCKER_REGISTRY}/eq-survey-runner-benchmark:${IMAGE_TAG} \
+        --set userWaitTimeMinSeconds=${USER_WAIT_TIME_MIN_SECONDS} \
+        --set userWaitTimemaxSeconds=${USER_WAIT_TIME_MAX_SECONDS} \
+        --set parallelism=${PARALLELISM} \
+        --set output.bucket=${GCS_OUTPUT_BUCKET} \
+        --set output.directory="timed-schedule/${RUNTIME_DATE_STRING}"
```
MebinAbraham: We should pass in the directory from concourse so it will be:
```yaml
--set output.directory="daily-test/${RUNTIME_DATE_STRING}"
```
or
```yaml
--set output.directory="stress-test/${RUNTIME_DATE_STRING}"
```

ci/run-benchmark.yaml
```diff
@@ -0,0 +1,60 @@
...
+      helm tiller run \
+        helm upgrade --install \
+        runner-benchmark \
+        k8s/helm \
+        --set requestsJson="${REQUESTS_JSON}" \
+        --set locustOptions="${LOCUST_OPTIONS}" \
+        --set host=https://${RUNNER_FULLY_QUALIFIED_DOMAIN_NAME} \
+        --set container.image=${DOCKER_REGISTRY}/eq-survey-runner-benchmark:${IMAGE_TAG} \
+        --set userWaitTimeMinSeconds=${USER_WAIT_TIME_MIN_SECONDS} \
+        --set userWaitTimemaxSeconds=${USER_WAIT_TIME_MAX_SECONDS} \
```
MebinAbraham:
```suggestion
        --set userWaitTimeMaxSeconds=${USER_WAIT_TIME_MAX_SECONDS} \
```

PR #18, opened 2020-01-28
Graphing benchmark results

berroar:
## Context
This PR has been raised to add graphical representation of the results generated by the daily benchmark tests. Scripts have been added to download the benchmark results from GCP, and graph the results.

## How to Review
Follow the steps in the Readme to download the benchmark results and check the graph that is generated.

See if you agree with the approach taken.


berroar:
[Graph eq-daily-performance-test results](https://trello.com/c/dxJJc54y/3592-graph-eq-daily-performance-test-results)

scripts/visualise_results.py
```diff
@@ -0,0 +1,80 @@
...
+    plt.xlabel("Run Date (DD-MM)")
+    plt.savefig('performance_graph.png')
+
+
+def get_run_date(folder):
+    for file in os.listdir(folder):
+        if "output_requests-metadata.json" in file:
+            with open(f'{folder}/{file}') as json_file:
+                metadata = json.load(json_file)
+                return datetime.utcfromtimestamp(int(metadata["timestamp"])).strftime('%d-%m')
```
ajmaddaford: The folder name contains the data and I think it would be simpler to use that.

scripts/get_benchmark_results.py
```diff
@@ -0,0 +1,23 @@
...
+if __name__ == '__main__':
+    gcs_bucket_name = os.environ['GCS_OUTPUT_BUCKET']
+
+    if not gcs_bucket_name:
+        raise ValueError(
+            'Value for `GCS_OUTPUT_BUCKET` environment variable not valid.'
+        )
+
+    application_credentials = os.environ['GOOGLE_APPLICATION_CREDENTIALS']
+    if not gcs_bucket_name:
```
MebinAbraham:
```suggestion
    if not application_credentials:
```

scripts/visualise_results.py
```diff
@@ -0,0 +1,80 @@
...
+        get_request_response_times = []
+        post_request_response_times = []
+        all_response_times = []
+
+        for file in os.listdir(folder):
+            if 'distribution.csv' not in file:
+                continue
+
+            with open(f'{os.getcwd()}/{folder}/{file}') as f:
+                data = f.read()
```
ajmaddaford: `glob` can be used to match filenames. This also has the advantage it returns the full file path, so it makes the `open` command clearer too.

```python
for file in glob(folder + '/*distribution.csv'):

    with open(file) as f:
        data = f.read()
```

scripts/get_benchmark_results.py
```diff
@@ -0,0 +1,23 @@
...
+        raise ValueError(
+            'Value for `GOOGLE_APPLICATION_CREDENTIALS` environment variable not valid.'
+        )
+
+    gcs = GoogleCloudStorage(bucket_name=gcs_bucket_name)
+    gcs.get_files()
```
MebinAbraham: It's' worth printing 'Fetching files' or something like that as there is a wait and at first I thought it wasn't running.

scripts/visualise_results.py
```diff
@@ -0,0 +1,80 @@
...
+                metadata = json.load(json_file)
+                return datetime.utcfromtimestamp(int(metadata["timestamp"])).strftime('%d-%m')
+
+
+if __name__ == '__main__':
+    test_run_folders = sorted(glob("outputs/daily-test/*"))
+
+    data_frame = get_stats(test_run_folders)
+
+    plot_data(data_frame)
```
MebinAbraham: Feedback after this would be handy.

README.md
```diff
@@ -163,6 +163,34 @@ helm tiller run \
...
+- Click on the `results-analysis` service account
+- Click edit, then, click 'Create Key'
+- Select JSON and store the file in a safe location - the path for this file will be used as an environment variable for the `get_benchmark_results.py` script.
+
+### Download Benchmark Results
+
+Run the `get_benchmark_results.py` to download the latest benchmark results from GCP. This script requires two environment variables to be set:
+
+`GCS_OUTPUT_BUCKET` - NAME OF THE STORAGE BUCKET
+`GOOGLE_APPLICATION_CREDENTIALS` - JSON KEYS FILE PATH FOR THE SERVICE ACCOUNT (see above)
```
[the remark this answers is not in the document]
MebinAbraham: @berroar You need to add `__init__.py` to your scripts folder. As per Andrew's comment the README should the be 
```bash
GCS_OUTPUT_BUCKET=eq-daily-performance-test-benchmark-outputs GOOGLE_APPLICATION_CREDENTIALS=<path_to_json_credentials_file> pipenv run python -m scripts.get_benchmark_results
```

Pipfile
```diff
@@ -14,9 +14,11 @@ sdc-cryptography = "==0.2.3"
...
 google-cloud-storage = "*"
+pandas = "*"
+matplotlib = "*"
 
 [requires]
 python_version = "3.7"
 
 [pipenv]
-allow_prereleases = false
+allow_prereleases = true
```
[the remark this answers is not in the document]
MebinAbraham: For: https://github.com/locustio/locust/blob/master/setup.py#L60

As discussed, we could look at the `FastHttpLocust` also.

scripts/get_benchmark_results.py
```diff
@@ -0,0 +1,23 @@
+import os
+
+from scripts.google_cloud_storage import GoogleCloudStorage
```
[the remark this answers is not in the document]
MebinAbraham: https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/18#discussion_r372827812 Should fix Andrew's comment @berroar

iwootten:
I don't think using pandas and matplotlib here are the right libraries for the job. It takes the current virtualenv from ~80MB to over 250MB. These are data science tools and seem overkill for graphing simple x, y values on a chart. I realise these are going into a dev environment, but ultimately we want this to be automated in a container somewhere.

I would have thought getting the data into a json format and pulling into a frontend lib like chartjs (169KB) would have been a better approach.

ajmaddaford:
I agree we should look at an alternative longer term solution, but this works and is providing value now. Also as the dependencies are development ones they aren't built into the container that is used to run the tests.

README.md
```diff
@@ -163,6 +163,46 @@ helm tiller run \
...
+```
+```
+GOOGLE_APPLICATION_CREDENTIALS - <path_to_json_credentials_file> pipenv run python -m scripts.get_benchmark_results
+```
+
+### Run the Visualise Results script
+
+The `visualise_results.py` script will run against any benchmark results stored in an `outputs/daily-test` directory.
+
+A line chart will be generated and saved as `performance_graph.png`.
```
ajmaddaford: This needs to be updated to reflect that the directory is now passed in. This should also have a run command and example specified. As the directory is passed in as an argument to the script here, we should consider doing the same for the bucket name in the get script above.

scripts/google_cloud_storage.py
```diff
@@ -0,0 +1,41 @@
...
+            if not os.path.exists(output_file_path):
+                os.makedirs(output_file_path)
+
+            blob.download_to_filename(filename=f'{output_dir}/{blob.name}')
+
+            metadata_file = (
+                f"{output_dir}/{file_path}/metadata.json"
+            )
+            with open(metadata_file, 'w+') as f:
+                f.write(json.dumps(blob.metadata, indent=2))
```
ajmaddaford: This will keep overwriting the metadata file - we should only write it once. We could assign the metadata to a variable and write it outside of the loop.

scripts/visualise_results.py
```diff
@@ -0,0 +1,76 @@
...
+    plt.yticks(size="small")
+    plt.ylabel("Average Response Time (ms)")
+    plt.xlabel("Run Date (DD-MM)")
+    plt.savefig('performance_graph.png')
+    print("Graph saved as performance_graph.png")
+
+
+if __name__ == '__main__':
+    if len(sys.argv) != 2:
+        print("Provide the benchmark outputs directory as a parameter")
```
ajmaddaford: This could be confusing as it needs the directory within `outputs`, adding "e.g. outputs/daily-test" would help.

scripts/visualise_results.py
```diff
@@ -0,0 +1,76 @@
...
+                    get_values.append(percentile_99th)
+                if 'POST /questionnaire' in line:
+                    post_values.append(percentile_99th)
+
+            get_request_response_times.extend(get_values)
+            post_request_response_times.extend(post_values)
+
+            all_response_times = get_values + post_values
+
+        date = folder.split('T')[0].split('/')[-1]
```
ajmaddaford: The split on `/` should be first, so it reads as "get the parent folder name and split on the date".

README.md
```diff
@@ -163,6 +163,46 @@ helm tiller run \
...
+
+### Pre-Requisites
+
+You need to be authenticated with GCP in order to download the benchmark results. To do this:
+
+Log In with gcloud auth using:
+
+`gcloud auth login`
+
+`gcloud auth application-default login`
```
ajmaddaford: I had to run `gcloud auth application-default login --project <valid_project_id>` or I got this error:

```
google.api_core.exceptions.BadRequest: 400 GET https://storage.googleapis.com/storage/v1/b/eq-daily-performance-test-benchmark-outputs?projection=noAcl: User project specified in the request is invalid.
```

ajmaddaford: So it looks like I was getting errors because the project I had set in the gcloud config was invalid (an old project). This can be set by:
- adding it to the config using - `gcloud config set project <project_id>`, or 
- passing it to the gcloud auth login command - `gcloud auth login --project <project_id>`

This needs to be set before running the `gcloud auth application-default login` command. I think we should change `gcloud auth login` to `gcloud auth login --project <project_id>` to prevent people from getting this error.

README.md
```diff
@@ -163,6 +163,46 @@ helm tiller run \
...
+### Download Benchmark Results
+
+Run the `get_benchmark_results.py` to download the latest benchmark results from GCP. This script requires two environment variables to be set:
+
+```
+GCS_OUTPUT_BUCKET - eq-daily-performance-test
+```
+```
+GOOGLE_APPLICATION_CREDENTIALS - <path_to_json_credentials_file> pipenv run python -m scripts.get_benchmark_results
+```
```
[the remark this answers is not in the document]
berroar: I've updated this section to show we only need this if running with the service account

scripts/google_cloud_storage.py
```diff
@@ -0,0 +1,43 @@
...
+
+            blob = self.bucket.blob(f'{directory or ""}/{output_filename}')
+            blob.metadata = {**kwargs}
+            blob.upload_from_filename(filename=output_file)
+
+    def get_files(self, output_dir):
+
+        for blob in self.client.list_blobs(self.bucket_name):
+
+            file_path = '/'.join(iter(blob.name.split('/')[:2]))
```
MebinAbraham: Built-in `rsplit` could be used.
```suggestion
            file_path = blob.name.rsplit('/', 1)[0]
```

scripts/get_benchmark_results.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+if __name__ == '__main__':
+    output_dir = "outputs"
```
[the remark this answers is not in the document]
berroar: I've now hardcoded this in the get files method

README.md
```diff
@@ -163,6 +163,55 @@ helm tiller run \
...
+
+`gcloud auth application-default login`
+
+Alternatively you can use the project service Account:
+
+- Go to the `eq-daily-performance-test` project in GCP
+- Go to IAM&Admin and open up the Service Accounts tab
+- Click on the `results-analysis` service account
+- Click edit, then, click 'Create Key'
+- Select JSON and store the file in a safe location - the path for this file will be used as an environment variable for the `get_benchmark_results.py` script.
```
MebinAbraham: This is specific to the project. I think all we need to say here is that you can use a service account which must have the necessary permission on the bucket.

README.md
```diff
@@ -163,6 +163,55 @@ helm tiller run \
...
+
+### Download Benchmark Results
+
+Run the `get_benchmark_results` script to download the latest benchmark results from GCP. You will need to set the bucket name as an environment variable:
+
+```
+GCS_OUTPUT_BUCKET - eq-daily-performance-test
```
MebinAbraham: Don't think this should use an actual bucket name. Maybe something like `The name of the bucket where the outputs are stored`

README.md
```diff
@@ -163,6 +163,55 @@ helm tiller run \
...
+```
```
ajmaddaford: If this was passed as a command line argument rather than an environment variable it would mean less documentation and it would be consistent with the visualise results script.

berroar: I had it like this because it was already being used as an environment variable for the store_benchmark_outputs.py script. Perhaps I should update that one too?

ajmaddaford: I think the get script should be consistent with the visualise script, and the store script left as is. It makes sense for store to use the environment variable as it's available in the container it runs in.

README.md
berroar: This line is no longer in the readme, changed from environment variable to cli  parameter.

PR #23, opened 2020-03-10
Better error handling for 200 and 302 responses

ajmaddaford:
# Context

Currently when something breaks in a request file it is hard to detect. This pull request enforces expected 200 responses for GET requests (a redirect would mean the requested page wasn't possible) and 302 for POST responses (a 200 response would probably mean a form validation failure). 

If an un-expected response is detected, the response is failed and an exception raised. Failing the response highlights the endpoint that caused the failure in the CSV file output. Raising the exception aborts the task, meaning no further requests for the task are executed after a failing one. It also gives clear failures when running from the command line.

# How to test

- Run the test requests files and verify the output
- Try an incorrect URL on a GET request and review the output
- Try an incorrect POST (e.g. invalid radio option) and review the output

### DIFF
--- a/requests/census_household_gb_eng.json
+++ b/requests/census_household_gb_eng.json
@@ -1,6 +1,10 @@
 {
     "requests": [
         {
+            "method": "GET",
+            "url": "/questionnaire/who-lives-here-interstitial/"
+        },
+        {
             "method": "POST",
             "url": "/questionnaire/who-lives-here-interstitial/",
             "data": {}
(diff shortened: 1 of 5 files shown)


ajmaddaford:
@LJBabbage The expected 404 returning 302 is a runner thing; for block urls runner validates the requested location and if not permitted it will redirect. In terms of the test requests files, any GET *should* be a 200.

The tests continue to loop after a failure, but for each task don't run any more requests from a requests file after a failed request.

PR #24, opened 2020-03-19
Update pip packages

petechd:
This updates the Pipfile packages, so we can use the latest version of locustio. It adds the latest locustio outputs: a new output_failures.csv file which will be stored in the GCS bucket when running benchmarks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ FROM python:3.7-slim
 WORKDIR /benchmark
 COPY . /benchmark
 
+RUN apt-get update && apt-get install -y gcc python3-dev
 # Install the required dependencies via pip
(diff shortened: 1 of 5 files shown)


petechd:
[Update eq-survey-runner-benchmark's Pip packages](https://trello.com/c/jioHUeht/3737-update-eq-survey-runner-benchmarks-pip-packages)

Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM python:3.7-slim
 WORKDIR /benchmark
 COPY . /benchmark
 
+RUN apt-get update && apt-get install -y gcc python3-dev
```
ajmaddaford: I think it would be worth adding a comment to say these dependencies are required for the psutil Python package.

PR #25, opened 2020-03-25
Migrate to github actions

petechd:
This moves build process from Travis CI to GitHub Actions to bring it in-line with other eq repos (runner, schemas and validator). There are two files: one for pull requests and one for master push. The only action required is docker build push, there is no linting or testing at the moment.


petechd:
[Migrate eq-survey-runner-benchmark from Travis to Github actions (S/M)](https://trello.com/c/9bH6E9hz/3751-migrate-eq-survey-runner-benchmark-from-travis-to-github-actions-s-m)

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,21 @@
+name: PR
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  docker-push:
+    runs-on: ubuntu-latest
```
berroar: Should these be pinned to version 18.04 in line with other repos?

.github/workflows/master.yml
```diff
@@ -0,0 +1,19 @@
+name: Master
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  docker-push:
+    runs-on: ubuntu-latest
```
JamesGardiner:
```suggestion
    runs-on: ubuntu-18.04
```

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,21 @@
+name: PR
+
+on:
+  pull_request:
+    branches:
+      - master
+
```
JamesGardiner: Should this specify master as the branch for pull requests?

petechd: I need to ask Ian about this one, I followed his code in validator repo but the schemas one doesn't specify master in branches.

.github/workflows/pull_request.yml
petechd: Will pin it now, thanks!

.github/workflows/master.yml
petechd: Thanks👍

.github/workflows/pull_request.yml
JamesGardiner: I believe this will only trigger the actions on pull requests on Master; we shouldn't restrict the running of the actions in that way. Very rarely do we pull into branches other than master, but when we do we should trigger this workflow.

petechd: ok, removing now👍

petechd:
> Is the push to GCR no longer needed?

It is, my mistake. I changed it to GCR now👍

.github/workflows/master.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build
+        run: docker build -t eu.gcr.io/census-eq-ci/eq-survey-runner-benchmark:latest .
+      - name: Push
+        run: |
+          echo "${{ secrets.DOCKER_GCR_PASSWORD }} | base64 --decode | docker login -u ${{ secrets.DOCKER_GCR_USERNAME }} --password-stdin https://eu.gcr.io
+          echo "Pushing with tag [latest]"
+          docker push eu.gcr.io/census-eq-ci/eq-survey-runner-benchmark:latest
```
MebinAbraham: Only comment would be is whether `eu.gcr.io/census-eq-ci` should be added as a:
`"${{ secrets.DOCKER_GCR_REGISTRY }}"`

@JamesGardiner Thoughts?

ajmaddaford: I think we should be consistent with other repos and push to Dockerhub rather than GCR.

petechd: Done!👍

PR #30, opened 2020-04-24
Use correct percentile in analysis

ajmaddaford:
Previous to this change the incorrect column was being used so we were reporting on the 99.9th percentile timings rather than the 99th.

I have also removed the logic for `_distribution.csv` files as they no longer exist and their replacement (`_output_stats_history.csv`) uses a different format.

### How to test
- Follow the instructions in the Readme to run the visualise results script with the contents of the `eq-daily-performance-test-benchmark-outputs` bucket
- Verify the correct timings are graphed

### DIFF
--- a/scripts/visualise_results.py
+++ b/scripts/visualise_results.py
@@ -15,7 +15,7 @@ def get_stats(folders):
         post_request_response_times = []
         all_response_times = []
 
-        for file in glob(folder + '/*distribution.csv') or glob(folder + '/*stats.csv'):
+        for file in glob(folder + '/*stats.csv'):
 
             with open(file) as f:
                 data = f.read()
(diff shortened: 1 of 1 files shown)


berroar:
Not sure if I've made a mistake here but after doing a fresh download with `pipenv run python -m scripts.get_benchmark_results eq-daily-performance-test-benchmark-outputs` the visualise script falls over because the download was still pulling down the _distribution.csv files and the visualise script can't work out the mean as a result.

ajmaddaford:
@berroar You're right, there is an issue with old data folders that contain any `_distribution.csv` files. Removing these folders (anything prior to 13th February) fixes the issue. I propose we remove these from the bucket rather than making the script work with deprecated files.

ajmaddaford:
@berroar Those output files are deleted now. If you remove your `outputs` directory and re-download it should work.

PR #34, opened 2020-05-06
Performance investigation for Python 3.8

pricem14pc:
Confirm we are happy with the testing carried out and the decision to merge the runner PR https://github.com/ONSdigital/eq-questionnaire-runner/pull/94


doc/performance-investigations/0004-update-to-python-38/summary.md
```diff
@@ -0,0 +1,47 @@
+# Update to Python 3.8
+
+Following gevent now supporting Python 3.8 (as of gevent 1.5) we are now able to upgrade Python.
+There will be performance implications with both Pythion 3.8 and dependent the gunicorn/gevent libraries.
```
JamesGardiner, line 4:
```suggestion
There will be performance implications with both Python 3.8 and the gunicorn/gevent libraries.
```

MebinAbraham: I've used `Pythion` before, no as good as `Python` tbh.

doc/performance-investigations/0004-update-to-python-38/summary.md
```diff
@@ -0,0 +1,47 @@
...
+| Questionnaire GETs | 587 | 393 |
+| Questionnaire POSTs | 730 | 431 |
+| All requests | 690 | 380 |
+
+## Supporting Observations
+![](0004-response-rates-and-latencies.png)
+
+## Decision
+
+Significant performance gains indicate a merge of this branch would be beneficial
```
JamesGardiner, line 47: Performance gains shouldn't be this easy :+1:

doc/performance-investigations/0004-update-to-python-38/summary.md
```diff
@@ -0,0 +1,47 @@
...
+Results based on 99th percentile timings.
+
+| Metric | Baseline | Investigation |
+|--------|----------|--------------|
+| Questionnaire GETs | 587 | 393 |
+| Questionnaire POSTs | 730 | 431 |
+| All requests | 690 | 380 |
+
+## Supporting Observations
+![](0004-response-rates-and-latencies.png)
```
JamesGardiner, line 43: This renders to nothing for me

pricem14pc: Refresh your browser. Is there another way to insert the image?

JamesGardiner: This is the url generated in the diff on Github https://github.com/ONSdigital/eq-survey-runner-benchmark/blob/master/doc/performance-investigations/0004-update-to-python-38/0004-response-rates-and-latencies.png - it's pointing to master not this branch, so I suspect it's a GitHub issue.

doc/performance-investigations/0004-update-to-python-38/summary.md
pricem14pc: Oh yes they should ;o)

pricem14pc:
I also did testing over multiple runs of the baseline and candidate at different times to what's on this PR (and later to consolidate these findings also with different configurations). The results varied run to run upto what's documented here. The results here were pretty much the results from the first stable runs following minimal change when deploying the app (essentially just the image change). Until we work out why there is variation run-to-run then IF these investigations need to include averages from multiple runs then we need to be clear about how that is done. I'm happy to close this PR and let Mebin document his findings. However, while maybe not quite hitting the results in this PR, I think the later configurations I have tested do suggest even more gains than Mebin's results at a higher load.

ajmaddaford:
I don't think it's about needing to include the average from multiple runs. If there are clear performance differences (in percentage terms) between different loads, then we should document them in this pull request. We could add the results from one run that Mebin did, and add a line about the performance differences being greater under heavier load?

pricem14pc:
Well the PR was raised prior to carrying more extensive testing under various loads. But I can add in the last one I did (which I've shared on the slack channel). The benchmark and candidate tests were following the same configuration as previous performance test investigations hence me posting only the results of that one test run.

PR #39, opened 2020-06-12
Allow non-minified html pages in csrf regex

iwootten:

Benchmark doesn't correctly identify non minified versions of pages, due to an extra forward slash at the end of the `<input />`. This change adds it to allow locust to match non-minified pages.

### How to review 

Run a journey locally against runner with the env var EQ_ENABLE_HTML_MINIFY set to True and then False and ensure both run through the journeys successfully.


iwootten:
[Performance investigation for removing minification of HTML output](https://trello.com/c/6SkOMFNy/3921-performance-investigation-for-removing-minification-of-html-output)

PR #41, opened 2020-06-29
Performance investigation for removing minification of HTML output

iwootten:

The latest daily test highlights the time taken to minify the HTML output as one of the top 5 in terms of proportion of the request time. Given that we gzip the response, the effect of HTML minification is minimal. We should investigate disabling/removing it.

Uses the PR https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/39 to allow non minified html in test client and https://github.com/ONSdigital/eq-questionnaire-runner/pull/148 to disable minification in runner and turn on whitespace removal.

### How to review 

Decide if you think we should merge.


iwootten:
[Performance investigation for removing minification of HTML output](https://trello.com/c/6SkOMFNy/3921-performance-investigation-for-removing-minification-of-html-output)

ajmaddaford:
I think that's enough of an improvement to merge the change and see it's effect on the daily test.

doc/performance-investigations/0006-use-non-minified-html/summary.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Results
+
+Results based on 99th percentile timings.
+
+| Metric | Baseline | Investigation (Uncompressed) | Investigation (Jinja Flags) |
+|--------|----------|--------------|
+| Questionnaire GETs | 360 | 269 | 263 |
+| Questionnaire POSTs | 446 | 377 | 320 |
+| All requests | 490 | 390 | 330 |
```
MebinAbraham, line 41: This isn't rendering properly, 2nd line is missing a column.

PR #54, opened 2020-08-07
Benchmark to include request weighting

LJBabbage:
Our benchmark currently doesn't include request weighting, this will distort the results as different endpoints can get different volumes of requests.

N.B - There are a number of other improvements that can be made to these results, and I will stick a card in tech session to discuss, but haven't looked to change them on this PR. They include not including session and thank you GETS in any of our figures and splitting up the results on known pinch points (i.e. grouping features rather than our current aggregating everything)

### How to review 
Run a test and check benchmark_stats returns the expected figures.

### DIFF
--- a/scripts/benchmark_stats.py
+++ b/scripts/benchmark_stats.py
@@ -13,15 +13,14 @@ class BenchmarkStats:
         self.weighted_post_requests: List[int] = []
         self.total_get_requests: int = 0
         self.total_post_requests: int = 0
-        self.total_requests: int = 0
         self.total_failures: int = 0
         self._process_file_data()
 
     def __str__(self):
         return (
-            f'Questionnaire GETs average: {int(self.average_get)}ms\n'
-            f'Questionnaire POSTs average: {int(self.average_post)}ms\n'
-            f'All requests average: {int(self.average_total)}ms\n'
+            f'Questionnaire GETs average: {int(self.average_weighted_get)}ms\n'
+            f'Questionnaire POSTs average: {int(self.average_weighted_post)}ms\n'
+            f'All requests average: {int(self.average_weighted_total)}ms\n'
             f'Total Requests: {int(self.total_requests):,}\n'
             f'Total Failures: {int(self.total_failures):,}\n'
             f'Error Percentage: {(round(self.error_percentage, 2))}%\n'
@@ -39,12 +38,11 @@ class BenchmarkStats:
                             self.weighted_get_requests.append(percentile_99th * request_count)
                             self.total_get_requests += request_count
                         elif row["Type"] == "POST":
-                            self.weighted_post_requests += request_count
-                            self.post_requests.append(percentile_99th * request_count)
+                            self.weighted_post_requests.append(percentile_99th * request_count)
+                            self.total_post_requests += request_count
 
                     if row["Name"] == "Aggregated":
                         failure_count = row.get("Failure Count") or row.get("# failures")
-                        self.total_requests += request_count
                         self.total_failures += int(failure_count)
 
     @property
@@ -61,8 +59,12 @@ class BenchmarkStats:
 
     @property
     def average_weighted_total(self):
-        return sum(self.weighted_get_requests + self.weighted_post_requests) / (self.total_post_requests + self.total_get_requests)
+        return sum(self.weighted_get_requests + self.weighted_post_requests) / self.total_post_requests
 
     @property
     def error_percentage(self):
         return (self.total_failures * 100) / self.total_requests
+
+    @property
+    def total_requests(self):
+        return self.total_get_requests + self.total_post_requests
--- a/scripts/benchmark_stats.py
+++ b/scripts/benchmark_stats.py
@@ -12,6 +12,8 @@ class BenchmarkStats:
 
         self.get_requests: List[int] = []
         self.post_requests: List[int] = []
+        self.total_get_requests = 0
+        self.total_post_requests = 0
         self.total_requests: int = 0
         self.total_failures: int = 0
         self._process_file_data()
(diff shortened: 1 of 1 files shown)


berroar:
I thought the percentile was changed deliberately from 99.9 to 99 as part of this [PR]?(https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/30)

LJBabbage:
> I thought the percentile was changed deliberately from 99.9 to 99 as part of this

Odd, I asked the author of that PR and confirmed I was moving it back to 99.9%. Will check

scripts/benchmark_stats.py
```diff
@@ -49,15 +54,15 @@ def files(self) -> List[str]:
 
     @property
     def average_get(self):
-        return mean(self.get_requests)
```
berroar, line 52: Can remove the import of mean now as it is no longer used.

scripts/benchmark_stats.py
```diff
@@ -31,16 +32,19 @@ def _process_file_data(self):
...
                         elif row["Type"] == "POST":
-                            self.post_requests.append(percentile_99th)
+                            self.total_post_requests += request_count
+                            self.post_requests.append(percentile_99th * request_count)
 
                     if row["Name"] == "Aggregated":
-                        request_count = row.get("Request Count") or row.get("# requests")
                         failure_count = row.get("Failure Count") or row.get("# failures")
-                        self.total_requests += int(request_count)
+                        self.total_requests += request_count
```
MebinAbraham, line 47: We should now remove this as we have the total get/post requests. This should also be removed from `__init__` and be made a property which is used by `average_total()` and `__str__()`. 
So something like:
```py
    @property
    def total_requests(self):
        return self.total_get_requests + self.total_post_requests
    
    @property
    def average_total(self):
        return sum(self.post_requests + self.get_requests)/self.total_requests
```
Might also be worth rounding to 2 decimals.

LJBabbage: Well you might think that, but how we have configured this means it removes the GET's for session and thank you, which I don't understand and why I didn't remove this. Kind of why I put a card in tech session as this repo needs some refining

scripts/benchmark_stats.py
```diff
@@ -49,15 +53,15 @@ def files(self) -> List[str]:
...
 
     @property
     def average_post(self):
-        return mean(self.post_requests)
+        return sum(self.post_requests)/self.total_post_requests
 
     @property
     def average_total(self):
-        return mean(self.get_requests + self.post_requests)
+        return sum(self.post_requests + self.get_requests)/(self.total_post_requests + self.total_get_requests)
```
JamesGardiner, line 64:
```suggestion
        return sum(self.post_requests + self.get_requests) / (self.total_post_requests + self.total_get_requests)
```

scripts/benchmark_stats.py
```diff
@@ -49,15 +53,15 @@ def files(self) -> List[str]:
 
     @property
     def average_get(self):
-        return mean(self.get_requests)
+        return sum(self.get_requests)/self.total_get_requests
 
     @property
     def average_post(self):
-        return mean(self.post_requests)
+        return sum(self.post_requests)/self.total_post_requests
```
JamesGardiner, line 60:
```suggestion
        return sum(self.post_requests) / self.total_post_requests
```

scripts/benchmark_stats.py
```diff
@@ -49,15 +53,15 @@ def files(self) -> List[str]:
 
     @property
     def average_get(self):
-        return mean(self.get_requests)
+        return sum(self.get_requests)/self.total_get_requests
```
JamesGardiner, line 56:
```suggestion
        return sum(self.get_requests) / self.total_get_requests
```

scripts/benchmark_stats.py
```diff
@@ -31,16 +32,19 @@ def _process_file_data(self):
             with open(file) as fp:
                 for row in DictReader(fp, delimiter=","):
                     percentile_99th = int(row["99%"])
+                    request_count = int(row.get("Request Count") or row.get("# requests"))
+
                     if "/questionnaire" in row['Name']:
                         if row["Type"] == "GET":
-                            self.get_requests.append(percentile_99th)
+                            self.get_requests.append(percentile_99th * request_count)
```
JamesGardiner, line 39: It would be worth renaming this now to something like `weighted_get_requests`

scripts/benchmark_stats.py
```diff
@@ -31,16 +32,19 @@ def _process_file_data(self):
...
+                            self.total_get_requests += request_count
                         elif row["Type"] == "POST":
-                            self.post_requests.append(percentile_99th)
+                            self.total_post_requests += request_count
+                            self.post_requests.append(percentile_99th * request_count)
```
JamesGardiner, line 43: Same as with the get requests. This should be renamed to `weighted_post_requests`

scripts/benchmark_stats.py
JamesGardiner: If we do round then consider just changing the display formatting rather than the value itself using the `{:0.2f}` formatting spec

scripts/benchmark_stats.py
```diff
@@ -12,6 +11,8 @@ def __init__(self, folder_paths: List[str]):
 
         self.get_requests: List[int] = []
         self.post_requests: List[int] = []
+        self.total_get_requests: int = 0
+        self.total_post_requests: int = 0
         self.total_requests: int = 0
```
JamesGardiner, line 16: This doesn't need to be in the init anymore, it can be a property of the class and use the total_get_requests and total_post_requests to calculate it:

```python
@property
def total_requests(self):
    return self.total_get_requests + self.total_post_requests
```

LJBabbage: Same comment as above for Mebin, we oddly exclude thank you and session, so if I did that the numbers are wrong. This PR was a quick fix to get the request count in with a card in tech to work out what this repo is really meant to do, as its a bit of a pickle at the moment

scripts/benchmark_stats.py
LJBabbage: I will put rounding on the tech card, as well as testing

PR #57, opened 2020-08-17
Include all endpoints in benchmark stats

MebinAbraham:
This PR includes all endpoints in our benchmark stats

### How to review 
- Ensure the stats include figures from all endpoints.
- Ensure graphing works as expected.


scripts/benchmark_stats.py
```diff
@@ -21,7 +21,7 @@ def __str__(self):
             f'Questionnaire GETs average: {int(self.average_weighted_get)}ms\n'
             f'Questionnaire POSTs average: {int(self.average_weighted_post)}ms\n'
             f'All requests average: {int(self.average_weighted_total)}ms\n'
-            f'Total Requests: {int(self.total_requests):,}\n'
+            f'Total Questionnaire Requests: {int(self.total_requests):,}\n'
             f'Total Failures: {int(self.total_failures):,}\n'
             f'Error Percentage: {(round(self.error_percentage, 2))}%\n'
```
ajmaddaford, line 26: I think we need to review these numbers - should they all be in the context of Questionnaire requests, or should we not segregate questionnaire requests from others?

Currently:
- The failure count is for all endpoints, which is misleading when following request timings and numbers for Questionnaire requests. 
- The error percentage is wrong as it's dividing the total number of failures across all endpoints by the number of Questionnaire requests (rather than all requests).

MebinAbraham: As discussed in tech session, the stats now take into account all endpoints.

MebinAbraham:
> It looks like you've run `black` on the scripts. In this case, they should be added to the `lint` command, and we should fail the `lint` if `black` indicates changes would be made

Don't want to muddy this PR, so https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/58

JamesGardiner (approved):

PR #63, opened 2020-10-09
Update to python 3.8

berroar:
Updates the python version to 3.8 (some libraries used by this repo are not yet compatible with 3.9 e.g. matplotlib)

### How to review 
Check that the benchmark process still runs successfully
Check that the python scripts in this repo still work


berroar:
[Python 3.9 (L)](https://trello.com/c/sGLihMaz/4198-python-39-l)

petechd (approved):

MebinAbraham (changes requested):
The output results to slack CI job uses: `google/cloud-sdk:slim` which doesn't support 3.8.

```
+ pipenv install --deploy
Warning: the environment variable LANG is not set!
We recommend setting this in ~/.profile (or equivalent) for proper expected behavior.
Warning: Python 3.8 was not found on your system…
Neither 'pyenv' nor 'asdf' could be found to install Python.
You can specify specific versions of Python with:
$ pipenv --python path/to/python
```

Using the non `slim` image worked.

.python-version
```diff
@@ -1 +1 @@
-3.7.6
+3.8.5
```
MebinAbraham, line 1: Any reason to now use 3.8.6?

berroar: I used 3.8.5 because I hadn't seen 3.8.6 in my available pyenv versions, updated it now though 👍

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8-alpine
+FROM python:3.8-buster
```
MebinAbraham, line 1: The python version in the dockerfile is fine. The issue is with cloud sdk in the ci task.
https://github.com/ONSdigital/eq-survey-runner-benchmark/blob/master/ci/output-results-to-slack.yaml#L5-L6

JamesGardiner: cloud sdk will only support up to 3.7 won't it?

JamesGardiner: As per the cloud-sdk docs : `Cloud SDK requires Python. Supported versions are 3.5 to 3.7, and 2.7.9 or higher.`
`

MebinAbraham: GCS supports 3.8. But I don't think we need the cloud sdk image for the ci task, just a python image.

Pipfile
```diff
@@ -17,9 +17,11 @@ google-cloud-storage = "*"
 pandas = "*"
 matplotlib = "*"
 slackclient = "*"
+pillow = "*"
+pyyaml = "*"
```
MebinAbraham, line 21: Were these explicitly installed due to dependency issues but since they are also using the latest version I don't get why these were added?

ci/output-results-to-slack.yaml
```diff
@@ -3,7 +3,7 @@ image_resource:
   type: docker-image
   source:
     repository: google/cloud-sdk
-    tag: slim
+    tag: latest
```
MebinAbraham, line 6: You should be able to use:
```
    repository: python
    tag: 3.8
```
It is a smaller image than gcloud-sdk latest plus we don't actually use need any of the cloud-sdk cli commands.

Dockerfile
JamesGardiner: Yeah sorry that's what I was getting at - we shouldn't need it just for the Slack publish notification task

MebinAbraham (approved):

PR #66, opened 2020-10-20
Summary and results for 0009-eq-session-datastore-over-redis

pricem14pc:
Performance investigation comparing v3.52.0 (eq-session in Redis) with a branch where eq-session is in Datastore.

Does the investigation, results and write up make sense, and is the decision to merge appropriate given these findings?


doc/performance-investigations/0009-eq-session-datastore-over-redis/summary.md
```diff
@@ -0,0 +1,58 @@
...
+| Metric | Baseline | Investigation |
+|--------|----------|--------------|
+| Questionnaire GETs | 333ms | 340ms|
+| Questionnaire POSTs | 374ms| 379ms|
+| All requests | 353ms | 359ms |
+| Total requests (over 20 mins) | 750,432 | 746,720 |
+
+Environment monitoring observations
+
+- 10 Runner instances (30 cores)
```
MebinAbraham, line 34: Is there a specific reason we chose to do 10 instances instead of mimicking the daily test set up?. As I thought we planned to do all these with similar structure for a better comparison.

doc/performance-investigations/0009-eq-session-datastore-over-redis/summary.md
```diff
@@ -0,0 +1,58 @@
...
+## Results
+
+Locust results based on 99th percentile timings.
+
+| Metric | Baseline | Investigation |
+|--------|----------|--------------|
+| Questionnaire GETs | 333ms | 340ms|
+| Questionnaire POSTs | 374ms| 379ms|
+| All requests | 353ms | 359ms |
+| Total requests (over 20 mins) | 750,432 | 746,720 |
```
MebinAbraham, line 30: I am not quite sure how with a max usage of up to 16 cores out of 30, the response time is so high. Would be good to know if 100th percentile is particularly high which may have inflated the response times?

pricem14pc: The 100% should be in the results folder csv files. I will double check and add them if we think it will be useful. We haven't previously.

> I am not quite sure how with a max usage of up to 16 cores out of 30, the response time is so high

Compared to what? We see the response times increase at about ~50% CPU so I don't think that is unexpected. Daily test is < ~50% CPU so it's not unexpected they will see lower latencies in the results. We are comparing the baseline to the investigation here, not to the daily tests. If we want to start doing that then we should run these tests in the eq-test-environment (which I actually suggested for the purposes of this particular investigation but the team decided that wasn't appropriate).

doc/performance-investigations/0009-eq-session-datastore-over-redis/summary.md
pricem14pc: The comparison is against a benchmark for any individual investigation. We shouldn't infer anything for trying to compare against previous investigations (or the daily test). In this particular case we discussed and agreed to use 10 instances as it would also give an indication of the impact at slightly higher loads. The load injectors were increased in order to maintain a similar (albeit slightly higher) volume of requests over this number of instances.

The difference in latencies is only part of the picture though. The amount of CPU to serve those requests (even at higher latencies) also increases and 10 instances was intended to give a better indication of what that impact was.

If we feel we should always run these in exactly the same config as the daily tests to be able to compare not only to the benchmark but also across investigations we can discuss that and capture it as a rule for future tests.

MebinAbraham (approved):

doc/performance-investigations/0009-eq-session-datastore-over-redis/summary.md
MebinAbraham: We should prob investigate response times at different CPU usages.

doc/performance-investigations/0009-eq-session-datastore-over-redis/summary.md
MebinAbraham: It wasn't to compare to existing benchmark but it was to make things more controlled between different tests, hence why we said we would use daily test set up for https://trello.com/c/XqqqZvuh/4217-gunicorn-performance-benchmark-m

As far as the investigation is concerned I agree the comparison is against the baseline and it is enough to highlight what to do next.

JamesGardiner (approved):

PR #68, opened 2020-11-03
Filter benchmark results download by number of days

berroar:
Adds the ability to filter the benchmark results download from the GCS bucket by number of days. This can be done by running the `get_benchmark_results` script with the optional `NUMBER_OF_DAYS` environment variable set.

### How to review 
Run the `get_benchmark_results` script with the `NUMBER_OF_DAYS` environment variable set e.g.

`OUTPUT_BUCKET="eq-daily-performance-test-benchmark-outputs" NUMBER_OF_DAYS=5  pipenv run python -m scripts.get_benchmark_results`

### DIFF
--- a/scripts/get_benchmark_results.py
+++ b/scripts/get_benchmark_results.py
@@ -15,7 +15,7 @@ if __name__ == '__main__':
     from_date = None
 
     if number_of_days:
-        from_date = (datetime.utcnow() - timedelta(days=number_of_days))
+        from_date = datetime.utcnow() - timedelta(days=number_of_days)
 
     if not output_bucket:
         print("'OUTPUT_BUCKET' environment variable must be provided")
--- a/scripts/get_benchmark_results.py
+++ b/scripts/get_benchmark_results.py
@@ -11,6 +11,7 @@ if __name__ == '__main__':
 
     parsed_variables = parse_environment_variables()
     number_of_days = parsed_variables['number_of_days']
+    output_dir = parsed_variables['output_dir']
 
     from_date = None
 
@@ -24,5 +25,5 @@ if __name__ == '__main__':
     gcs = GoogleCloudStorage(bucket_name=output_bucket)
     print("Fetching files...")
 
-    gcs.get_files(from_date=from_date)
+    gcs.get_files(from_date=from_date, output_dir=output_dir)
     print('All files downloaded')
--- a/scripts/get_benchmark_results.py
+++ b/scripts/get_benchmark_results.py
@@ -14,10 +14,11 @@ if __name__ == '__main__':
     number_of_days = parsed_variables['number_of_days']
     output_dir = parsed_variables['output_dir']
 
-    from_date = None
-
-    if number_of_days:
-        from_date = datetime.now(tz=tzutc()) - timedelta(days=number_of_days)
+    from_date = (
+        datetime.now(tz=tzutc()) - timedelta(days=number_of_days)
+        if number_of_days
+        else None
+    )
 
     if not output_bucket:
         print("'OUTPUT_BUCKET' environment variable must be provided")
--- a/scripts/get_benchmark_results.py
+++ b/scripts/get_benchmark_results.py
@@ -1,17 +1,14 @@
 import os
 import sys
 
+from scripts.get_summary import parse_environment_variables
 from scripts.google_cloud_storage import GoogleCloudStorage
 
 if __name__ == '__main__':
     output_bucket = os.getenv("OUTPUT_BUCKET")
 
-    days = os.getenv("NUMBER_OF_DAYS")
-    if days and days.isdigit() is False:
-        print("'NUMBER_OF_DAYS' environment variable must be a valid integer value")
-        sys.exit(1)
-
-    days = int(days) if days else None
+    parsed_variables = parse_environment_variables()
+    number_of_days = parsed_variables['number_of_days']
 
     if not output_bucket:
         print("'OUTPUT_BUCKET' environment variable must be provided")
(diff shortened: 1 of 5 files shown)


berroar:
[Update eq-survey-runner-benchmark to only download required files](https://trello.com/c/ohhlaEmy/4272-update-eq-survey-runner-benchmark-to-only-download-required-files)

scripts/google_cloud_storage.py
```diff
@@ -24,10 +25,20 @@ def upload_files(
...
             blob.upload_from_filename(filename=output_file)
 
-    def get_files(self):
+    def get_files(self, number_of_days):
         output_dir = "outputs"
 
         for blob in self.client.list_blobs(self.bucket_name):
+            blob_date = blob.name.split("/")[1].split("T")[0]
+
+            from_date = (
```
LJBabbage, line 34: This doesn't need to be in the loop

scripts/google_cloud_storage.py
```diff
@@ -24,10 +25,20 @@ def upload_files(
             blob.metadata = {**kwargs}
             blob.upload_from_filename(filename=output_file)
 
-    def get_files(self):
+    def get_files(self, number_of_days):
```
LJBabbage, line 28: Might be worth considering just passing the from date reducing the logic in the method

scripts/get_benchmark_results.py
```diff
@@ -6,12 +6,19 @@
 if __name__ == '__main__':
     output_bucket = os.getenv("OUTPUT_BUCKET")
 
+    days = os.getenv("NUMBER_OF_DAYS")
+    if days and days.isdigit() is False:
+        print("'NUMBER_OF_DAYS' environment variable must be a valid integer value")
+        sys.exit(1)
+
+    days = int(days) if days else None
```
LJBabbage, line 14: With the other comment in mind this might be easier
```
from_date = None
number_of_days = os.getenv("NUMBER_OF_DAYS")

    if number_of_days:
        if not isinstance(number_of_days, int):
            print("'NUMBER_OF_DAYS' environment variable must be a valid integer value")
            sys.exit(1)
        from_date = (datetime.utcnow() - timedelta(days=number_of_days))
```

LJBabbage (changes requested):
A couple of minor points. Although on the card not a huge fan of "number_of_days" var name, its not hugely clear what that means

scripts/get_benchmark_results.py
MebinAbraham, line 14: This should just reuse the existing method we have for that handles this env var.
```py
parsed_variables = parse_environment_variables()
number_of_days = parsed_variables['number_of_days']
```

berroar: I've added this, although now it makes the `OUTPUT_DIR` env mandatory for this script

berroar: I've moved the logic, and also simplified this a bit by using the existing `parse_environment_variables` method that @MebinAbraham commented about below

scripts/google_cloud_storage.py
```diff
@@ -24,10 +25,14 @@ def upload_files(
             blob.metadata = {**kwargs}
             blob.upload_from_filename(filename=output_file)
 
-    def get_files(self):
-        output_dir = "outputs"
+    def get_files(self, from_date):
+        output_dir = os.getenv("OUTPUT_DIR")
```
MebinAbraham, line 29: This should be passed in.

scripts/get_benchmark_results.py
```diff
@@ -1,17 +1,28 @@
...
 
+from datetime import datetime, timedelta
+
+from scripts.get_summary import parse_environment_variables
 from scripts.google_cloud_storage import GoogleCloudStorage
 
 if __name__ == '__main__':
     output_bucket = os.getenv("OUTPUT_BUCKET")
 
+    parsed_variables = parse_environment_variables()
```
MebinAbraham, line 12: I think it is fine to default output dir to `outputs` as part of `parse_environment_variables` then pass it to `get_files()`.

LJBabbage (changes requested):
Looks good and works as expect, the only thing that needs doing is the Readme needs updating.

scripts/google_cloud_storage.py
```diff
@@ -24,10 +25,13 @@ def upload_files(
...
             blob.upload_from_filename(filename=output_file)
 
-    def get_files(self):
-        output_dir = "outputs"
+    def get_files(self, from_date, output_dir):
 
         for blob in self.client.list_blobs(self.bucket_name):
+            blob_date = blob.name.split("/")[1].split("T")[0]
+
+            if from_date and datetime.strptime(blob_date, "%Y-%m-%d") < from_date:
```
MebinAbraham, line 33: Blob object already have timestamps so this can be:
```suggestion
            if from_date and blob.time_created < from_date:
```
However, you will need to make sure `from_date` is time zone aware as you can't compare timezone aware and timezone naive dates.
Like so:
```py
from_date = datetime.now(tz=tzutc()) - timedelta(days=number_of_days)
```
tzutc can be imported using: `from dateutil.tz import tzutc` as it is already part of our dependency.

LJBabbage (approved):

MebinAbraham (changes requested):
The CI task for outputting results to Slack need to be updated to include the dir and number of days.
```
      # Get benchmark outputs
      OUTPUT_BUCKET="$OUTPUT_BUCKET" \
      OUTPUT_DIR="outputs" \
      NUMBER_OF_DAYS="$NUMBER_OF_DAYS" \
      pipenv run python -m scripts.get_benchmark_results
```

MebinAbraham (commented):
Just doing a final test but formal CI being super slow.

MebinAbraham (approved):

PR #69, opened 2020-11-04
Performance Investigation: Gunicorn async and threaded web servers

MebinAbraham:
Adds the finding from carrying out the test of async vs threaded web servers under Gunicorn.

### How to review 
Is the document clear and highlight what it needs to?
You will have to view the file by clicking "View file" to see any embedded images or [click here](https://github.com/ONSdigital/eq-survey-runner-benchmark/blob/performance-investigation-async-threads/doc/performance-investigations/0010-web-server-worker-class/summary.md)


doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
+# Web server worker class: Async vs Threads vs Threads + GRPC
+
+We have only ever run the application with Gunicorn in async mode (workers). We should try threads if it is more performant, reliable or consistent.
```
LJBabbage, line 3: Past doesn't really mean much or provide a justification for a test, specially if you are reading in the future, as context is lost . This is simply a Gunicorn in async mode compared to it using threads with and without gRPC

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
...
+This test replicates the current settings for the daily test.
+
+The version of runner tested: https://github.com/ONSdigital/eq-questionnaire-runner/releases/tag/v3.53.0
+
+## Results
+
+The tests were run multiple times to ensure the performance was consistent; however, I have only included a single sample per configuration in the outputs.
+The results below are the averages between all tests. Note response times are in milliseconds.
+
+|                          | Request Count | Failure Count | Median Response Time | Average Response Time | Min Response Time | Max Response Time | Average Content Size | Requests/s |
```
LJBabbage, line 30: With my maths hat on, you shouldn't mix median and average response time, it should be median and mean in this scenario as median is a classification of average

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
+# Web server worker class: Async vs Threads vs Threads + GRPC
```
LJBabbage, line 1: Isn't it gRPC?

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
...
+![](outputs/performance_comparison.png)
+
+## Observations
+
+- Threads with GRPC is just as performant as async
+- Using threads is only feasible when also using GRPC
+
+## Decision
+- Given the performance of Gunicorn threads with GRPC is on par with async gevent and the options it opens up, we should use threads with GRPC
+- Look at the impact of varying the number of threads
```
LJBabbage, line 61: Probably warrants an action heading

LJBabbage (changes requested):
Looks good, some minor stuff

doc/performance-investigations/0010-web-server-worker-class/summary.md
MebinAbraham: Just following our past pattern. But have updated.

doc/performance-investigations/0010-web-server-worker-class/summary.md
MebinAbraham: I agree but kept it as it is for clarity as this is what locust reports and what we have used in past reports.

doc/performance-investigations/0010-web-server-worker-class/summary.md
MebinAbraham: Can change but still also a decision?. Just trying to keep it consistent with other reports. Not to bothered either way.

pricem14pc:
While you didn't encounter any errors under HTTP during these tests, I wonder if it's worth mentioning the benefit of using GRPC over HTTP should mean we are no longer subject to the request connection exceptions that we have observed across previous tests? I think the next stress test will confirm if this is true.

MebinAbraham:
> While you didn't encounter any errors under HTTP during these tests, I wonder if it's worth mentioning the benefit of using GRPC over HTTP should mean we are no longer subject to the request connection exceptions that we have observed across previous tests? I think the next stress test will confirm if this is true.

I guess we don't know that yet to document it.

LJBabbage (approved):

pricem14pc (approved):
I think that given we have introduced the ability to change the web server configuration and the Datastore protocol accepting these changes (to threaded/GRCP) is low risk allowing us to easily test at scale and revert if necessary. The fact that threads is comparable performance to the current async config is probably tipped in favour of threads by the potential to use the formal Google libraries.

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
...
+
+Performance comparison of Async vs Thread with gRPC over multiple tests.
+
+I didn't include threads without gRPC since it was not performant enough to be considered.
+
+![](outputs/performance_comparison.png)
+
+## Observations
+
+- Threads with gRPC is just as performant as async
```
ajmaddaford, line 56: Until we run a scale test, we can't say that threads+grpc is as performant as async. Changing `is` to `was` focusses the observation on what was tested.

```suggestion
- Threads with gRPC was as performant as async
```

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
...
+| Gunicorn Threads         | 60     | 67  | 72  | 77  | 92     | 110 | 140 | 160 | 250    | 450    | 1500 |
+| Gunicorn Threads + gRPC  | 50     | 55  | 59     | 62  | 72  | 83  | 100 | 120 | 190   | 350   | 770  |
+
+|                          | GETs Response Time | POSTs Response Time| All requests Response Time | 
+|--------------------------|--------------------|--------------------|----------------------------|
+| Gunicorn Async           | 87                 | 111                | 99                        |
+| Gunicorn Threads         | 137                | 162                | 150                        |
+| Gunicorn Threads + gRPC  | 91                 | 110                | 100                         |
+
+Performance comparison of Async vs Thread with gRPC over multiple tests.
```
ajmaddaford, line 48: - It's not clear what this means, are the numbers above not for multiple tests? 
- If this is associated with the image, think it would be clearer as a heading

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
...
+
+I didn't include threads without gRPC since it was not performant enough to be considered.
```
ajmaddaford, line 50: "I didn't" doesn't fit with the rest of the document. It should be something like:
```suggestion
Threads without gRPC is not included as it was not performant enough.
```

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Observations
+
+- Threads with gRPC is just as performant as async
+- Using threads is only feasible when also using gRPC
+
+## Decision
+- Given the performance of Gunicorn threads with gRPC is on par with async gevent and the options it opens up, we should use threads with gRPC
+- Look at the impact of varying the number of threads
+- Test uWSGI threads with gRPC
```
ajmaddaford, line 62: We should add another action to run a scale test.

doc/performance-investigations/0010-web-server-worker-class/summary.md
```diff
@@ -0,0 +1,62 @@
+# Web server worker class: Async vs Threads vs Threads + gRPC
```
ajmaddaford, line 1: We have previously tested async vs threads vs threads + gRPC in https://github.com/ONSdigital/eq-survey-runner-benchmark/blob/master/doc/performance-investigations/0008-use-gthread-grpc/summary.md. I think we should present this as a re-run of that investigation now that sessions are in Datastore. By linking to that investigation all of the documentation around the why of threads and gRPC is already covered.

doc/performance-investigations/0010-web-server-worker-class/summary.md
MebinAbraham: Made it a heading.

doc/performance-investigations/0010-web-server-worker-class/summary.md
MebinAbraham: Done.

doc/performance-investigations/0010-use-gthread-grpc-2/summary.md
```diff
@@ -0,0 +1,61 @@
...
+| Gunicorn Threads         | 60     | 67  | 72  | 77  | 92     | 110 | 140 | 160 | 250    | 450    | 1500 |
+| Gunicorn Threads + gRPC  | 50     | 55  | 59     | 62  | 72  | 83  | 100 | 120 | 190   | 350   | 770  |
+
+|                          | GETs Response Time | POSTs Response Time| All requests Response Time | 
+|--------------------------|--------------------|--------------------|----------------------------|
+| Gunicorn Async           | 87                 | 111                | 99                        |
+| Gunicorn Threads         | 137                | 162                | 150                        |
+| Gunicorn Threads + gRPC  | 91                 | 110                | 100                         |
+
+#### Performance comparison of Async vs Thread with gRPC over multiple tests.
```
ajmaddaford, line 46:
```suggestion
#### Performance comparison of Async vs Thread with gRPC over multiple tests
```

doc/performance-investigations/0010-use-gthread-grpc-2/summary.md
```diff
@@ -0,0 +1,61 @@
...
+|--------------------------|-----|-----|-----|-----|-----|-----|-----|-----|--------|--------|------|
+| Gunicorn Async           | 52     | 58  | 61     | 64  | 72  | 82  | 97  | 120 | 200    | 350    | 720  |
+| Gunicorn Threads         | 60     | 67  | 72  | 77  | 92     | 110 | 140 | 160 | 250    | 450    | 1500 |
+| Gunicorn Threads + gRPC  | 50     | 55  | 59     | 62  | 72  | 83  | 100 | 120 | 190   | 350   | 770  |
+
+|                          | GETs Response Time | POSTs Response Time| All requests Response Time | 
+|--------------------------|--------------------|--------------------|----------------------------|
+| Gunicorn Async           | 87                 | 111                | 99                        |
+| Gunicorn Threads         | 137                | 162                | 150                        |
+| Gunicorn Threads + gRPC  | 91                 | 110                | 100                         |
```
ajmaddaford, line 44: Given some effort has gone in to line up the table headings with the lines, the same should be done for the table contents.

ajmaddaford (approved):

doc/performance-investigations/0010-use-gthread-grpc-2/summary.md
MebinAbraham: Had it perfectly aligned -_- think I ran some sort of formatter which messed it up 😞. Have fixed.

doc/performance-investigations/0010-use-gthread-grpc-2/summary.md
MebinAbraham: Fixed

PR #73, opened 2020-11-30
Graph percentiles

MebinAbraham:
Updates the performance graph to use percentiles.

The percentiles, as well as GET/POST averages are still included as part of the summary.

### How to review 
Ensure:
 - graph plots the percentiles
 - summary outputs percentiles as well as existing statistics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,25 +216,24 @@ pipenv run python -m scripts.get_summary
 
 This will output something like:
 ```
+2020-01-01
 ---
 Percentiles:
-50th: 74ms
-90th: 140ms
-95th: 170ms
-99th: 240ms
-99.9th: 360ms
-100th: 820ms
+50th: 92ms
+90th: 252ms
+95th: 320ms
+99th: 467ms
+99.9th: 664ms
+100th: 1179ms
 ---
-GETs (99th): 237ms
-POSTs (99th): 272ms
+GETs (99th): 466ms
+POSTs (99th): 467ms
 ---
-Total Requests: 323,148
+Total Requests: 307,207
 Total Failures: 0
 Error Percentage: 0.0%
-```
 
-*Note: The values for GETs (99th) and POSTs (99th) will not match up with the 99th percentile because GET/POST values are calculated on a per endpoint basis whereas the
-percentile outputs are from the `Aggregated` row of the locust output file.*
+```
 
 If `OUTPUT_DATE` is not provided, then it will output a summary for all results within the provided directory.
 
@@ -250,14 +249,14 @@ This will output something like:
 ---
 Percentiles:
 50th: 76ms
-90th: 149ms
-95th: 182ms
-99th: 263ms
-99.9th: 407ms
-100th: 1425ms
+90th: 147ms
+95th: 177ms
+99th: 254ms
+99.9th: 372ms
+100th: 484ms
 ---
-GETs (99th): 237ms
-POSTs (99th): 272ms
+GETs (99th): 235ms
+POSTs (99th): 273ms
 ---
 Total Requests: 3,209,821
 Total Failures: 0
--- a/README.md
+++ b/README.md
@@ -216,31 +216,52 @@ pipenv run python -m scripts.get_summary
 
 This will output something like:
 ```
-Questionnaire GETs average: 477ms
-Questionnaire POSTs average: 573ms
-All requests average: 524ms
-Total Requests: 209303
+---
+Percentiles:
+50th: 74ms
+90th: 140ms
+95th: 170ms
+99th: 240ms
+99.9th: 360ms
+100th: 820ms
+---
+GETs (99th): 237ms
+POSTs (99th): 272ms
+---
+Total Requests: 323,148
 Total Failures: 0
 Error Percentage: 0.0%
 ```
 
+*Note: The values for GETs (99th) and POSTs (99th) will not match up with the 99th percentile because GET/POST values are calculated on a per endpoint basis whereas the
+percentile outputs are from the `Aggregated` row of the locust output file.*
+
 If `OUTPUT_DATE` is not provided, then it will output a summary for all results within the provided directory.
 
 ### Summarise the Stress Test results:
-To get a breakdown of results for a stress test use the `get_stress_test_summary` script. This accepts a folder containing
+To get a breakdown of results for a stress test use the `get_aggregated_summary` script. This accepts a folder containing
 results as a parameter, and will provide aggregate totals at the folder level:
 ```bash
-OUTPUT_DIR="outputs/stress-test" pipenv run python -m scripts.get_stress_test_summary
+OUTPUT_DIR="outputs/stress-test" pipenv run python -m scripts.get_aOggregated_summary
 ```
 
 This will output something like:
 ```
-Questionnaire GETs average: 465ms
-Questionnaire POSTs average: 525ms
-All requests average: 495ms
-Total Requests: 30056523
-Total Failures: 140186
-Error Percentage: 0.47%
+---
+Percentiles:
+50th: 76ms
+90th: 149ms
+95th: 182ms
+99th: 263ms
+99.9th: 407ms
+100th: 1425ms
+---
+GETs (99th): 237ms
+POSTs (99th): 272ms
+---
+Total Requests: 3,209,821
+Total Failures: 0
+Error Percentage: 0.0%
 ```
 
 ### Run the Visualise Results script
--- a/README.md
+++ b/README.md
@@ -218,13 +218,12 @@ This will output something like:
 ```
 2020-01-01
 ---
-Percentiles:
+Percentile Averages:
 50th: 92ms
 90th: 252ms
 95th: 320ms
 99th: 467ms
 99.9th: 664ms
-100th: 1179ms
 ---
 GETs (99th): 466ms
 POSTs (99th): 467ms
@@ -247,13 +246,12 @@ OUTPUT_DIR="outputs/stress-test" pipenv run python -m scripts.get_aggregated_sum
 This will output something like:
 ```
 ---
-Percentiles:
+Percentile Averages:
 50th: 76ms
 90th: 147ms
 95th: 177ms
 99th: 254ms
 99.9th: 372ms
-100th: 484ms
 ---
 GETs (99th): 235ms
 POSTs (99th): 273ms
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ If `OUTPUT_DATE` is not provided, then it will output a summary for all results 
 To get a breakdown of results for a stress test use the `get_aggregated_summary` script. This accepts a folder containing
 results as a parameter, and will provide aggregate totals at the folder level:
 ```bash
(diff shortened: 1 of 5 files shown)


scripts/visualise_results.py
```diff
@@ -34,16 +35,21 @@ def plot_data(df, number_of_days_to_plot):
...
-            result[1].average_weighted_get,
-            result[1].average_weighted_post,
-            result[1].average_weighted_total,
+            *(
+                result[1].percentiles[percentile]
+                for percentile in BenchmarkStats.PERCENTILES_TO_GRAPH
+            ),
         ]
         for result in results
     ]
```
MebinAbraham, line 48: If this is not clear, happy to update to:
```py
    result_fields = []
    for result in results:
        percentiles = (
            result[1].percentiles[percentile]
            for percentile in BenchmarkStats.PERCENTILES_TO_GRAPH
        )
        result_fields.append([result[0], *percentiles])
```

LJBabbage (changes requested):
I found this output confusing

```
Percentiles:
50th: 60ms
90th: 85ms
95th: 98ms
99th: 140ms
99.9th: 240ms
100th: 850ms
---
GETs average: 106ms
POSTs average: 135ms
All requests average: 121ms
---
Total Requests: 204,701
Total Failures: 0
Error Percentage: 0%

```
I don't understand what the 2nd chunk of outputs is or refers to or has any relationship to the percentiles. If it is the 99th, then the average does need to line up.

MebinAbraham:
> I found this output confusing
> 
> ```
> Percentiles:
> 50th: 60ms
> 90th: 85ms
> 95th: 98ms
> 99th: 140ms
> 99.9th: 240ms
> 100th: 850ms
> ---
> GETs average: 106ms
> POSTs average: 135ms
> All requests average: 121ms
> ---
> Total Requests: 204,701
> Total Failures: 0
> Error Percentage: 0%
> ```
> 
> I don't understand what the 2nd chunk of outputs is or refers to or has any relationship to the percentiles. If it is the 99th, then the average does need to line up.

I'm not sure we need the 2nd chunk tbh. I kept it as it already existed. I'll leave it open for discussion. 

The main difference is the 2nd chunk is the weighted 99th percentile but given our journey is consistent between routes, and now that we report all percentiles, when it was discussed, the impression was that we don't need weighting. I will leave it open for further discussion whether we want to keep this, or we could consider weighting all. If we want to keep, then I think we should either weight everything or don't weight given our benchmark is consistent for most routes.

scripts/benchmark_stats.py
```diff
@@ -1,77 +1,95 @@
+from collections import defaultdict
 from csv import DictReader
 from glob import glob
-from typing import List
+from typing import List, Mapping
 
 
 class BenchmarkStats:
+    PERCENTILES_TO_REPORT = (50, 90, 95, 99, 99.9, 100)
+    PERCENTILES_TO_GRAPH = PERCENTILES_TO_REPORT[:-2]
```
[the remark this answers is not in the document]
MebinAbraham: Done

MebinAbraham:
@LJBabbage As discussed in tech session, I have removed the `All requests average` and removed the weighting from GET/POST requests.

scripts/benchmark_stats.py
```diff
@@ -1,77 +1,84 @@
...
-            f'GETs average: {int(self.average_weighted_get)}ms\n'
-            f'POSTs average: {int(self.average_weighted_post)}ms\n'
-            f'All requests average: {int(self.average_weighted_total)}ms\n'
-            f'Total Requests: {int(self._total_requests):,}\n'
-            f'Total Failures: {int(self._total_failures):,}\n'
+            f'---\n'
+            f'Percentiles:\n'
+            f'{formatted_percentiles}\n'
+            f'---\n'
+            f'GETs average: {self.average_get}ms\n'
```
LJBabbage, line 33: Probably still needs the Heading 99%. Although this still looks weird on the results I am using, as both GET and POST are lower then in the percentiles block
```
---
Percentiles:
50th: 60ms
90th: 85ms
95th: 98ms
99th: 140ms
99.9th: 240ms
100th: 850ms
---
GETs average: 108ms
POSTs average: 131ms
---
Total Requests: 204,701
Total Failures: 0
Error Percentage: 0%
```

LJBabbage (changes requested):

scripts/benchmark_stats.py
```diff
@@ -1,77 +1,84 @@
...
-            / self._total_requests
-        )
+    def average_post(self) -> int:
+        return int(mean(self._post_requests))
 
     @property
-    def error_percentage(self):
-        return (self._total_failures * 100) / self._total_requests
+    def error_percentage(self) -> int:
+        return (self._total_failures * 100) // self._total_requests
```
LJBabbage, line 84: Double divide?

LJBabbage (changes requested):

scripts/benchmark_stats.py
MebinAbraham: No, it is Python's floor division operator. However, I have removed it as we want the granularity of the error percentages as a floating point as floor division will remove that given the rate of error should be very low.

scripts/benchmark_stats.py
pricem14pc: As discussed, I think it's still worth outputting POST/GET statistics separately (particularly now we've removed that split from the graph) as although historically they have tracked each other, it might highlight future issues. Adding the fact they're 99th and clarifying somewhere in the code/readme why they wont straddle the 99th in the above table is probably good enough

scripts/visualise_results.py
```diff
@@ -34,16 +35,21 @@ def plot_data(df, number_of_days_to_plot):
...
     result_fields = [
         [
             result[0],
-            result[1].average_weighted_get,
-            result[1].average_weighted_post,
-            result[1].average_weighted_total,
+            *(
+                result[1].percentiles[percentile]
+                for percentile in BenchmarkStats.PERCENTILES_TO_GRAPH
+            ),
```
ajmaddaford, line 45: `results` is a `NamedTuple` with `date` and `statistics` fields, so `result[0]` can be replaced with `result.date` and `result[1]` with `result.statistics`.

scripts/visualise_results.py
```diff
@@ -34,16 +35,21 @@ def plot_data(df, number_of_days_to_plot):
...
         ]
         for result in results
     ]
 
-    data_frame = DataFrame(result_fields, columns=["DATE", "GET", "POST", "AVERAGE"])
+    percentile_columns = (
+        f"{percentile}th" for percentile in BenchmarkStats.PERCENTILES_TO_GRAPH
```
ajmaddaford, line 51: I think the value for `PERCENTILES_TO_GRAPH` should be defined in this file, as it is only used in this file.

scripts/benchmark_stats.py
```diff
@@ -1,77 +1,84 @@
...
 from glob import glob
-from typing import List
+from statistics import mean
+from typing import List, Mapping
 
 
 class BenchmarkStats:
+    PERCENTILES_TO_REPORT = (50, 90, 95, 99, 99.9, 100)
+    PERCENTILES_TO_GRAPH = (50, 90, 95, 99)
+    PERCENTILE_TO_USE_FOR_ROUTES = 99
```
ajmaddaford, line 11: I'm not sure about the naming here, maybe `PERCENTILE_TO_USE_FOR_AVERAGES`?

scripts/benchmark_stats.py
MebinAbraham: Have updated the output and readme. Take a look and let me know if it is sufficient.

scripts/visualise_results.py
MebinAbraham: Good shout!. Updated.

README.md
```diff
@@ -216,31 +216,52 @@ pipenv run python -m scripts.get_summary
...
+99th: 263ms
+99.9th: 407ms
+100th: 1425ms
+---
+GETs (99th): 237ms
+POSTs (99th): 272ms
+---
+Total Requests: 3,209,821
+Total Failures: 0
+Error Percentage: 0.0%
```
ajmaddaford, line 264: The aggregated percentiles could be misleading when the number of requests is different in each folder being summarised. It would work ok for something like the daily test where the number of requests is constant, but I think when used for something like a stress test where we ramp up, the numbers will be lower than they should be.

README.md
```diff
@@ -216,31 +216,52 @@ pipenv run python -m scripts.get_summary
...
+
 If `OUTPUT_DATE` is not provided, then it will output a summary for all results within the provided directory.
 
 ### Summarise the Stress Test results:
-To get a breakdown of results for a stress test use the `get_stress_test_summary` script. This accepts a folder containing
+To get a breakdown of results for a stress test use the `get_aggregated_summary` script. This accepts a folder containing
 results as a parameter, and will provide aggregate totals at the folder level:
 ```bash
-OUTPUT_DIR="outputs/stress-test" pipenv run python -m scripts.get_stress_test_summary
+OUTPUT_DIR="outputs/stress-test" pipenv run python -m scripts.get_aOggregated_summary
```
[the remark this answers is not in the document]
MebinAbraham: Fixed, cheers.

README.md
```diff
@@ -216,31 +216,52 @@ pipenv run python -m scripts.get_summary
...
+GETs (99th): 237ms
+POSTs (99th): 272ms
+---
+Total Requests: 323,148
 Total Failures: 0
 Error Percentage: 0.0%
 ```
 
+*Note: The values for GETs (99th) and POSTs (99th) will not match up with the 99th percentile because GET/POST values are calculated on a per endpoint basis whereas the
+percentile outputs are from the `Aggregated` row of the locust output file.*
```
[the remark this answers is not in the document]
MebinAbraham: This has now been removed because everything is now weighted.

README.md
MebinAbraham: All response times are now weighted.

LJBabbage (approved):
Looks much better

doc/performance-investigations/README.md
```diff
@@ -63,10 +63,18 @@ OUTPUT_DIR=outputs/baseline pipenv run python -m scripts.get_aggregated_summary
...
-POSTs average: 146ms
-All requests average: 131ms
-Total Requests: 204,438
+Percentiles:
+50th: 74ms
+90th: 140ms
+95th: 170ms
+99th: 240ms
+99.9th: 360ms
+100th: 820ms
```
ajmaddaford, line 72: I noticed that the 100th percentile was quite a bit lower with the recent changes for weighting. After reading through https://www.circonus.com/2018/11/the-problem-with-percentiles-aggregation-brings-aggravation/ it looks like aggregating percentiles should not usually be done, as it doesn’t work as expected and hides outliers. However, as we don't have the original data, we can't do it properly, and the average percentiles are useful as an indication of performance. I think we should label them as `Percentile Averages` and remove the 100th percentile from the output as I think most people would expect that to be the longest request time.

MebinAbraham: Agreed. Have made the changes.

ajmaddaford (approved):

PR #78, opened 2020-12-23
Individual Response Performance test investigation

berroar:

Adds the performance investigation for individidual response endpoints: `/doc/performance-investigations/0014-individual-response-performance-test`

In order to carry out the test I had to amend the  `utils.py` to handle vars in URL query params as our benchmark app could not handle vars in URL query parameters e.g. `/individual-response/{person_1_list_id}/text/confirm-number?mobile_number={mobile_number}`.

### How to review 
Check the findings in the performance investigation document.

To test the change to the `utils.py`, use the newly added `individual_response` or `confirmation_email` journeys and make sure the benchmark is working as expected.


berroar:
[Performance test IR](https://trello.com/c/aVNBOp80/4221-performance-test-ir)

runner_benchmark/utils.py
```diff
@@ -18,4 +18,10 @@ def parse_params_from_location(url, route):
             if field_pos == value_pos:
                 params[field_name] = value
 
+    if urlparse(route).query:
+        query_result = urlparse(route).query.partition("=")[2]
+        url_query = urlparse(url).query.partition("=")[2]
+
+        params[query_result.strip("{}")] = url_query
+
```
ajmaddaford, line 26: - `urlparse` should be called once for each url
- this only works for one query parameter, we should look at using the `parse_qs` method from `urlparse` and make it work for any number of query parameters.

requests/individual_response.json
```diff
@@ -0,0 +1,219 @@
...
+                "individual-response-text-confirm-answer": "Yes, send the text"
+            }
+        },
+        {
+            "method": "GET",
+            "url": "/individual-response/text/confirmation?mobile_number={mobile_id}"
+        },
+        {
+            "method": "POST",
+            "url": "/individual-response/text/confirmation?mobile_number={mobile_id}",
```
ajmaddaford, line 152: Could we use `mobile_number` rather than `mobile_id`, or does that not work?

requests/confirmation_email.json
```diff
@@ -0,0 +1,191 @@
...
+            "method": "GET",
+            "url": "/submitted/thank-you/"
+        },
+        {
+            "method": "POST",
+            "url": "/submitted/thank-you/",
+            "data": {
+                "email": "example@example.com"
+            },
+            "redirect_route": "/submitted/confirmation-email/sent?email={email_id}"
```
ajmaddaford, line 33: Could we use `email` rather than `email_id`, or does that not work?

requests/confirmation_email.json
```diff
@@ -0,0 +1,191 @@
+{
```
MebinAbraham, line 1: Was this meant to be committed?

doc/performance-investigations/0014-individual-response-performance-test/outputs/readme.md
```diff
@@ -0,0 +1,52 @@
...
+| GET | /individual-response/post/confirmation | 82 |
+| POST | /individual-response/post/confirmation | 78 |
+| GET | /individual-response/text/confirmation?mobile_number={id} | 79 |
+| POST | /individual-response/text/confirmation?mobile_number={id} | 74 |
+| GET | /individual-response/{id}/how | 91 |
+| POST | /individual-response/{id}/how | 82 |
+| GET | /individual-response/{id}/post/confirm-address | 89 |
+| POST | /individual-response/{id}/post/confirm-address | 148 |
+| GET | /individual-response/{id}/text/confirm-number?mobile_number={id} | 87 |
+| POST | /individual-response/{id}/text/confirm-number?mobile_number={id} | 148 |
```
MebinAbraham, line 44: Worth updating "id" based on your recent changes.

requests/confirmation_email.json
berroar: I committed this file to help test the change to utils.py, as I wanted to make sure it worked with both the mobile number and email address, but can remove it if needed

ajmaddaford: I don't see why not. I asked Pete to commit the feedback one.

MebinAbraham: Ah makes sense.

runner_benchmark/utils.py
berroar: I've refactored this to use the parse_qs method now but think it still might be able to be shortened as wasn't sure what to change for your first point?

ajmaddaford: - `urlparse` is already called on lines 5 and 6. Those lines should be changed to store the parsed url, something like:

    ```python
    parsed_route = urlparse(route)
    parsed_url = urlparse(url)
    ```

    Then the path can be accessed with `parsed_route.path` and the query with `parsed_url.query`.

- When looping over the query params, can this be simplified to something like:

    ```python
    params.update((key, url_result[key][0]) for key in query_result)
    ```

- I think we might be able to use `parse_qsl` rather than `parse_qs` and remove the `[0]` access.

ajmaddaford (approved):

doc/performance-investigations/0014-individual-response-performance-test/outputs/readme.md
```diff
@@ -0,0 +1,52 @@
+# Individual Response Performance Investigation
```
MebinAbraham, line 1: Can this file be renamed to `summary.md` to be consistent with the exists docs?

requests/confirmation_email.json
MebinAbraham, line 1: Can you double check this?. I ran this against staging and got:
```
GET,/submitted/confirmation-email/send,Expected a (200) but got a (302) back when getting page: /submitted/confirmation-email/send,7
```

berroar: As discussed, removed a post in the requests file as my local env for the email limit had been set to 11 not 10.

MebinAbraham (approved):

PR #79, opened 2021-01-12
Add gunicorn thread config at scale investigation

ajmaddaford:
Documents an investigation into reducing the number of Gunicorn threads used by runner.

### How to review 
- Is there enough information
- Are there any other observations
- Do you agree with the decisions


MebinAbraham (approved):
Will be interesting to see the performance under fewer workers.

pricem14pc (approved):
I think this backs up the observations from the previous investigation and the expectation that the number of Redis connections correlates with the number of threads.
As discussed, a full scale test will give even further confidence around the appropriateness of 3 threads.

PR #80, opened 2021-01-19
Add Email confirmation and Feedback to household journey

LJBabbage:
This PR adds a single email confirmation and a single feedback submission to the household benchmark journey.

### How to review 
Run the benchmark script against your local or a GCP environment and make sure it works (pipenv run ./run.sh requests/census_household_gb_eng.json)

### DIFF
--- a/requests/census_household_gb_eng.json
+++ b/requests/census_household_gb_eng.json
@@ -1282,6 +1282,35 @@
         {
             "method": "GET",
             "url": "/submitted/thank-you/"
+        },
+        {
+            "method": "POST",
+            "url": "/submitted/thank-you/",
+            "data": {
+                "email": "test@email.com"
+            },
+            "redirect_route": "/submitted/confirmation-email/sent?email={email}"
+        },
+        {
+            "method": "GET",
+            "url": "/submitted/confirmation-email/sent?email={email}"
+        },
+        {
+            "method": "GET",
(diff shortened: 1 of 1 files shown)


requests/census_household_gb_eng.json
```diff
@@ -1282,6 +1282,35 @@
         {
             "method": "GET",
             "url": "/submitted/thank-you/"
+        },
+        {
+            "method": "POST",
+            "url": "/submitted/thank-you/",
+            "data": {
+                "email": "test@email.com"
```
MebinAbraham, line 1290: We prob need to use a specific email address for Gov Notify when used for test purposes?

LJBabbage: That's an interesting point, should test be hitting Notify? I am not sure, but then we have a problem with how we manage the cloud function. Easiest thing to do might be to change the email to a Notify one and decide later. It also opens up the question how we handle confirmation emails locally now. Realistically we don't want to hand the key around

LJBabbage: There might been some value in having the smoke test email for Notify, even if we do decide not to use it in the daily benchmark.

MebinAbraham (approved):

petechd (approved):
Tested this on my benchmark pipeline, works fine!👍

Screenshot 2021-01-20 at 10 24 49

PR #85, opened 2021-04-09
Update python and packages

JamesGardiner:
Updates the Python version to 3.9.4, and updates packages where outdated


petechd (commented):
We might have to upgrade sdc-cryptography to 0.3.0 so its dependencies will be unpinned and pyyaml will get updated to version 5.4.1 (otherwise it causes problems in `pipenv install --dev` as pyyaml 3.13 is depreciated).

JamesGardiner:
> We might have to upgrade sdc-cryptography to 0.3.0 so its dependencies will be unpinned and pyyaml will get updated to version 5.4.1 (otherwise it causes problems in `pipenv install --dev` as pyyaml 3.13 is depreciated).

I've unpinned the sdc-cryptography library 👍🏻

MebinAbraham (approved):

petechd (approved):

PR #86, opened 2021-04-14
Update Python 3.8 refs to 3.9

JamesGardiner:
Updates references to the wrong version of python (from 3.8 to 3.9).

### How to review 
Run the branch via a concourse pipeline, making sure the pipeline works as expected (particularly the publish to Slack)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8-slim
+FROM python:3.9-slim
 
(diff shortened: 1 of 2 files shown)


MebinAbraham (approved):
Both working as expected. 👍 

We should look to reduce the Docker image size in future as it is currently ~8GB 🙈

berroar (approved):

PR #88, opened 2021-05-18
Update test schema request JSON + remove census JSON

MebinAbraham:
Removed the Census schema request JSON files and updated the existing test schema one to use the `/submit` endpoint introduced by https://github.com/ONSdigital/eq-questionnaire-runner/pull/606

### How to review 
Run the benchmark using the updated requests file against the runner branch and ensure they work as expected.


berroar (changes requested):
The `run` command in the makefile needs to be updated as it still uses the census schema

MebinAbraham:
> The `run` command in the makefile needs to be updated as it still uses the census schema

Done

berroar (approved):

JamesGardiner (approved):

PR #93, opened 2021-09-16
Update Slack authorisation token

petechd:
This updates Slack authorisation token to our non-census specific one, used in the new Concourse, to communicate between pipelines and slack.

### How to review 
Check if you can post alerts to `eq-dev-team` or similar channel from your pipeline.


MebinAbraham (approved):
Approving on the basis that the key name has been updated correctly. Not checked whether it does in fact output a result.

berroar (approved):

MebinAbraham (changes requested):
Can you update the commit message to be more meaningful.

MebinAbraham (approved):

PR #94, opened 2021-09-17
Add method fix after google-cloud update

petechd:
This fixes a small bug that was introduced after `google-cloud` libraries update. `get_files` method stopped downloading results for the graph since the method started searching for `blobs` in top level directory.

### How to review 
Check [this PR in eq-pipelines](https://github.com/ONSdigital/eq-pipelines/pull/270) that makes use of the fix branch or run `get_benchmark_results` method locally.

### DIFF
--- a/scripts/google_cloud_storage.py
+++ b/scripts/google_cloud_storage.py
@@ -26,15 +26,9 @@ class GoogleCloudStorage:
 
     def get_files(self, from_date, output_dir):
 
-        blobs = (
-            blob
-            for blob in self.client.list_blobs(self.bucket_name)
-            if from_date
-            and blob.time_created >= from_date
-            and not blob.name.endswith("/")
-        )
-
-        for blob in blobs:
+        for blob in self.client.list_blobs(self.bucket_name):
+            if blob.name.endswith("/") or (from_date and blob.time_created < from_date):
+                continue
 
             file_path = blob.name.rsplit('/', 1)[0]
 
--- a/scripts/google_cloud_storage.py
+++ b/scripts/google_cloud_storage.py
@@ -26,13 +26,13 @@ class GoogleCloudStorage:
 
     def get_files(self, from_date, output_dir):
 
-        blobs = [
+        blobs = (
             blob
             for blob in self.client.list_blobs(self.bucket_name)
             if from_date
             and blob.time_created >= from_date
             and not blob.name.endswith("/")
-        ]
+        )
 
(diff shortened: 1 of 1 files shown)


petechd:
[Introduce EQ daily performance test (5 SP)](https://trello.com/c/wBIuJkpL/4869-introduce-eq-daily-performance-test-5-sp)

scripts/google_cloud_storage.py
```diff
@@ -28,7 +28,7 @@ def get_files(self, from_date, output_dir):
 
         for blob in self.client.list_blobs(self.bucket_name):
 
-            if from_date and blob.time_created < from_date:
+            if from_date and blob.time_created < from_date or blob.name.endswith("/"):
```
LJBabbage, line 31: This looks odd, and I found it hard to comprehend due to the mixing of types (i. date comparison and str checking) Can this not be done in the for loop?

petechd: So that `blob.name.endswith("/")` check is moved higher up one level where I create list of blobs, no such check in `for` loop anymore. Thanks!

scripts/google_cloud_storage.py
```diff
@@ -26,7 +26,13 @@ def upload_files(
 
     def get_files(self, from_date, output_dir):
 
-        for blob in self.client.list_blobs(self.bucket_name):
+        blobs = [
+            blob
+            for blob in self.client.list_blobs(self.bucket_name)
+            if not blob.name.endswith("/")
+        ]
```
MebinAbraham, line 33: I think it was fine as it was, but if you are using a list comprehension than the date comparison should be done here.

MebinAbraham: Also, you should also use a generator instead of a list comp in this scenario, as at the moment, you are looping over most of the matching elements twice.

petechd: I've just added date comparison conditions to list comprehension but will make it generator now, thanks!

petechd: Ok, added generator now, thanks!

petechd: Generator removed now as requested👍

MebinAbraham (approved):

LJBabbage (approved):

scripts/google_cloud_storage.py
petechd: it's all been moved around as we discussed! thanks!

PR #96, opened 2021-12-02
Add buiness benchmark request file

LJBabbage:
This PR add the request file for business. There are only 2 routing rules in the questionnaire and I have answered so every block is visited. I have also included a trip to the feedback page.

### How to review 
You will need runner to be using this branch, add-benchmark-business-schema  https://github.com/ONSdigital/eq-questionnaire-runner/pull/730. Then its a case of just running a test and making sure everything works as expected
```
pipenv run ./run.sh requests/test_benchmark_business.json
```

### DIFF
--- a/requests/test_benchmark_business.json
+++ b/requests/test_benchmark_business.json
@@ -1,0 +1,130 @@
+{
+    "requests": [
+        {
+            "method": "GET",
+            "url": "/questionnaire/introduction-block/"
+        },
+        {
+            "method": "POST",
+            "url": "/questionnaire/introduction-block/",
+            "data": {}
+        },
+        {
+            "method": "GET",
+            "url": "/questionnaire/block379/"
+        },
+        {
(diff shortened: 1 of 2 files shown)


petechd (approved):
I would use 'ref_p_start_date': '2019-05-01' and 'ref_p_end_date': '2019-08-01', since these are the dates that we use, but otherwise it LGTM and aggregates without a fail.

LJBabbage:
> I would use 'ref_p_start_date': '2019-05-01' and 'ref_p_end_date': '2019-08-01', since these are the dates that we use, but otherwise it LGTM and aggregates without a fail.

I didn't change the token values, just took away the rouge '0' The post date just fits the schema requirement

berroar (approved):
Journey appears to be ok 👍

MebinAbraham (approved):

PR #101, opened 2022-01-28
Remove output bucket var in run_benchmark script

petechd:
We wanted to hard code the output_bucket name on creation to "${var.project_id}-benchmark-outputs" in performance testing pipelines so we will pass the Project ID in future only.
Current approach is error prone and can lead to hard to diagnose test's timeouts.

### How to review 
Run one of the benchmarks using [appropriate branch](https://github.com/ONSdigital/eq-pipelines/pull/329) of  eq-pipelines and this branch of benchmark and check if test doesn't time out and outputs to correct bucket, slack channel.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -23,7 +23,6 @@ The following env vars must be set:
 - PARALLELISM
 - TIMEOUT
 - OUTPUT_DIR
-- OUTPUT_BUCKET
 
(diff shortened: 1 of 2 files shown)


petechd:
[Remove output bucket benchmark var from daily, performance test pipelines (1 SP)](https://trello.com/c/uFOPIgSc/5095-remove-output-bucket-benchmark-var-from-daily-performance-test-pipelines-1-sp)

MebinAbraham (approved):

berroar (approved):

PR #107, opened 2022-04-13
Ability to launch benchmark by schema url

berroar:

Runner can load a schema with a survey_url included in the JWT launch claim. This PR allows the benchmark to include this claim in the create_token method. Previously we only supported the schema_name.

### How to review 

Set the new `INCLUDE_SCHEMA_URL_IN_TOKEN` env var and run the benchmark using the `test_benchmark_business` schema.


berroar:
[Benchmark to launch by survey_url arg (2 SP)](https://trello.com/c/BUd5hCG9/5238-benchmark-to-launch-by-surveyurl-arg-2-sp)

petechd (changes requested):
The problem I found is: when you hit the `Schema url not present in the request file` exception it still outputs to the bucket with zero values and crashes `Output results to slack` job as a result.

Screenshot 2022-05-03 at 15 11 56

requests/test_benchmark_business.json
```diff
@@ -134,5 +134,6 @@
             "url": "/submitted/thank-you/"
         }
     ],
-    "schema_name": "test_benchmark_business"
+    "schema_name": "test_benchmark_business",
+    "schema_url": "https://storage.googleapis.com/eq-questionnaire-schemas/test_benchmark_business.json"
```
MebinAbraham, line 138: I think it is worth adding this key value into all our existing schemas in `requests/` folder.

berroar: Only benchmark business exists in that eq-questionnaire-schemas bucket though?

MebinAbraham: That's true, but our current scripts would be adding this key value in regardless of whether it exists in the bucket.

petechd (approved):

k8s/helm/templates/locust-job.yaml
```diff
@@ -31,6 +31,8 @@ spec:
               value: "{{ .Values.host }}"
```
MebinAbraham, line 31: Slightly unrelated to this change but `image: eu.gcr.io/census-eq-ci/eq-survey-runner-benchmark:latest` is wrong and is worth updating while we are here. Pipelines run fine since it overrides it. Worth pointing this to our dev gar. There is reference in the README to census gcr also that should be updated.

MebinAbraham (commented):
The `run.sh` script needs to be updated so that this can be tested locally also.

run.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 echo "Running Locust benchmark"
-USER_WAIT_TIME_MIN_SECONDS=0 USER_WAIT_TIME_MAX_SECONDS=0 REQUESTS_JSON="${1}" HOST="${2:-http://localhost:5000}" pipenv run locust --headless -u 1 -r 1 -t ${3:-1m} --csv=${4:-output} -L WARNING
+USER_WAIT_TIME_MIN_SECONDS=0 USER_WAIT_TIME_MAX_SECONDS=0 INCLUDE_SCHEMA_URL_IN_TOKEN="true" REQUESTS_JSON="${1}" HOST="${2:-http://localhost:5000}" pipenv run locust --headless -u 1 -r 1 -t ${3:-1m} --csv=${4:-output} -L WARNING
```
MebinAbraham, line 4: This should be configurable, like REQUEST_JSON and others are.

runner_benchmark/taskset.py
```diff
@@ -75,7 +78,10 @@ def handle_redirect(self, request, response):
         )
 
     def do_launch_survey(self):
-        token = create_token(schema_name=self.schema_name)
+        extra_payload = (
+            dict(schema_url=self.schema_url) if self.include_schema_url_in_token else {}
```
MebinAbraham, line 82: `self.schema_url` still needs to be set. Errors current since it doesn't exist.

run.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 echo "Running Locust benchmark"
-USER_WAIT_TIME_MIN_SECONDS=0 USER_WAIT_TIME_MAX_SECONDS=0 REQUESTS_JSON="${1}" HOST="${2:-http://localhost:5000}" pipenv run locust --headless -u 1 -r 1 -t ${3:-1m} --csv=${4:-output} -L WARNING
+USER_WAIT_TIME_MIN_SECONDS=0 USER_WAIT_TIME_MAX_SECONDS=0 REQUESTS_JSON="${1}" INCLUDE_SCHEMA_URL_IN_TOKEN="${2}" HOST="${3:-http://localhost:5000}"  pipenv run locust --headless -u 1 -r 1 -t ${4:-1m} --csv=${5:-output} -L WARNING
```
MebinAbraham, line 4: The README doesn't match this, given the order of arguments have changed. 2nd arg was the host but is now a bool.

MebinAbraham (approved):

PR #110, opened 2022-06-01
Performance test launching by schema url

berroar:

Adds performance investigation docs relating to launching schema's via a schema url rather than by schema name. 

### How to review 
Check the docs added


petechd (approved):

rmccar (approved):

MebinAbraham (commented):
Interesting findings. Maybe the spikes we see are due to the way cloud run manages the OS file system which isn’t real, it’s a in-memory file system, so should have been fast. Maybe once we do a stress test with schema_url, more info would surface.

PR #129, opened 2023-04-11
Set response_expires_at field to Payload

Yuyuutsu:
This PR is to fix the benchmark failures that are occurring because datastore jobs are removing "live" entities. We are adding the `response_expires_at` field which is defaulted value is 7 days from the current time to solve the issue.

### How to review
1. Run benchmark locally
2. Then deploy it on concourse and check if any errors occur

Is the way the time is calculated and formatted understandable?

### DIFF
--- a/runner_benchmark/token_generator.py
+++ b/runner_benchmark/token_generator.py
@@ -1,7 +1,7 @@
 import os
 import time
 from uuid import uuid4
-
+from datetime import datetime, timedelta
 from sdc.crypto.encrypter import encrypt
 from sdc.crypto.key_store import KeyStore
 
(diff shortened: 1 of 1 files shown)


runner_benchmark/token_generator.py
```diff
@@ -1,7 +1,7 @@
 import os
 import time
 from uuid import uuid4
-
+from datetime import datetime, timedelta
```
MebinAbraham, line 4:
```suggestion
from datetime import datetime, timedelta, timezone
```

petechd (approved):
Functionally tested, works fine!

PR #136, opened 2023-09-12
Update benchmark journey to use complex scenarios

Yuyuutsu:
Updated the business benchmark json and created a happy and unhappy json path.

This PR is linked with the PR in  [eq-questionnaire-runner](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1203).

[Trello Card](https://trello.com/c/RoMIXq3K/5634-update-benchmark-journey-to-test-eq-under-complex-scenarios-2-sp)
### How to review
Check if schema addresses all the new suggested scenarios in the Trello Card
Test both benchmarks locally and via a deployed staging 
The tag should be [update-benchmark-journey-to-test-complex-scenarios](https://console.cloud.google.com/artifacts/docker/ons-eq-ci/europe-west2/docker-images/eq-survey-runner-benchmark?pli=1&rapt=AEjHL4N0bDLdNykD9GD0Fvu11a1Jf4QHYJjMsgF_3C5iT4wiLrm3J5u-e5ehK7VWpV7nSygQZaW2fDg4_dVmVwnlN9qRFCCsvA)


Yuyuutsu:
[Update benchmark journey to test eQ under complex scenarios (2 SP)](https://trello.com/c/RoMIXq3K/5634-update-benchmark-journey-to-test-eq-under-complex-scenarios-2-sp)

Yuyuutsu:
> The existing request file `test_benchmark_business` will no longer work so should probably be removed. There are also some request files relating to the PDF download tests that will be broken by updating the new schema, so not sure if we want to keep these?

I shall remove the test_benchmark_business and update the PDF download tests we have 👍

MebinAbraham:
> The existing request file `test_benchmark_business` will no longer work so should probably be removed. There are also some request files relating to the PDF download tests that will be broken by updating the new schema, so not sure if we want to keep these?

We should keep any existing feature tests such as testing PDF download.

petechd:
Shouldn't we restore schemas that test pdf kit and weasy?

Yuyuutsu:
> Shouldn't we restore schemas that test pdf kit and weasy?

I have now pushed up PDF kit optimised 👍 . But weasy was removed because it was never meant to be in main.

MebinAbraham:
> > Shouldn't we restore schemas that test pdf kit and weasy?
> 
> I have now pushed up PDF kit optimised 👍 . But weasy was removed because it was never meant to be in main.

Both are not needed.

You can exclude PDF endpoint as they will skew the latency as PDF is significantly slower than the rest of the app.
A PDF one can be when we do PDF specific testing.

Yuyuutsu:
> > > Shouldn't we restore schemas that test pdf kit and weasy?
> > 
> > 
> > I have now pushed up PDF kit optimised 👍 . But weasy was removed because it was never meant to be in main.
> 
> Both are not needed.
> 
> You can exclude PDF endpoint as they will skew the latency as PDF is significantly slower than the rest of the app. A PDF one can be added when we do PDF specific testing.

Now removed 🫡

berroar (commented):
A PR will need to be raised for the pipelines repo if we are changing the name of the `test_benchmark_buisness` requests file, as that is the one referenced in the daily benchmark vars file

berroar (commented):
Possibly not relevant to this PR, but since we're updating the requests files here, the feedback requests json doesn't seem to work

berroar (commented):
The updated journeys/requests files are all working, but was the plan to update the benchmark scripts so that locust runs both the unhappy and happy paths?

requests/test_benchmark_business_pdfkit.json
```diff
@@ -11,94 +11,525 @@
         },
```
MebinAbraham, line 11: This file is not needed

MebinAbraham:
> The updated journeys/requests files are all working, but was the plan to update the benchmark scripts so that locust runs both the unhappy and happy paths?

Not yet, we should update to use happy path for daily test. We will do another card to do a comparison benchmark of the two. Then look to run unhappy only/both.

requests/test_benchmark_business_pdfkit.json
Yuyuutsu: Removed 👍

Yuyuutsu:
> Possibly not relevant to this PR, but since we're updating the requests files here, the feedback requests json doesn't seem to work

the schema changed but we forgot to change the feedback har file. I have a fix ready to push up if we want to push it up in this PR or in another PR

berroar:
> > Possibly not relevant to this PR, but since we're updating the requests files here, the feedback requests json doesn't seem to work
> 
> the schema changed but we forgot to change the feedback har file. I have a fix ready to push up if we want to push it up in this PR or in another PR

I'd create another PR for it 👍

petechd (approved):

PR #144, opened 2023-12-14
Update unhappy path remove-company endpoint POST

liamtoozer:
As a result of[ the bug fix](https://github.com/ONSdigital/eq-questionnaire-validator/pull/228) to prevent duplicate `answer_id` from appearing in different list collectors, the `test_benchmark_business.json` needed to be updated. This then meant that the benchmark journey for the unhappy path needed to be updated to pick up the answer_id change

### How to review
Run the unhappy path benchmark locally to ensure that the requests run correctly. 

**Note**: There is an existing issue with the unhappy path benchmark atm around the `GET,/questionnaire/people/{id}/set-min-max-block/` end point - this will be addressed in another ticket


liamtoozer:
[Bug: Validator allows duplicate answer ids in different list collectors](https://trello.com/c/fB5R54hy/5965-bug-validator-allows-duplicate-answer-ids-in-different-list-collectors)

berroar (approved):
Only get the pre-existing error 👍

PR #145, opened 2024-01-03
Fix unhappy path benchmark schema journey

VirajP1002:
The unhappy benchmark journey when run against main was where the following error was reported on the set-min-max-block page: Expected a (200) but got a (404) back when getting page: /questionnaire/people/{person_2}/set-min-max-block/?resume=True').

To fix this, a similar journey was followed to recreate an updated requests files and when running the benchmark tests locally, the error message does not appear.

### How to review
Run the benchmark tests using `make run` ensuring the file in the command is changed from `test_checkbox.json` to `test_benchmark_business_unhappy_path.json`


VirajP1002:
[Fix unhappy path benchmark schema journey](https://trello.com/c/SAGNqaey/5987-fix-unhappy-path-benchmark-schema-journey)

requests/test_benchmark_business_unhappy_path.json
```diff
@@ -155,7 +148,7 @@
             "url": "/questionnaire/people/add-person/",
             "data": {
                 "first-name": "Erling",
-                "last-name": "Haland"
+                "last-name": "Haaland"
```
berroar, line 151: 😆

berroar (approved):
Works, LGTM! 👍

petechd (approved):
Journey fixed, all good👍 We should make a new release after this merges?

requests/test_benchmark_business_unhappy_path.json
VirajP1002: Had to fix it! 😆

PR #146, opened 2024-01-15
Migrate from docker-image to registry-image

berroar:
Updates tasks to use `registry-image` rather than `docker-image`

### How to review
Deploy the benchmark pipeline and check that it is still working


berroar:
[Migrate benchmark tasks to registry-image](https://trello.com/c/yrQ9byJh/5999-migrate-benchmark-tasks-to-registry-image)

rmccar (approved):

PR #147, opened 2024-02-08
Add unit tests for the benchmark repo

berroar:

Adds unit tests for our scripts in our benchmark repo, including a mock output directory. The benchmark repo was initially created without any tests, so it is difficult to gain 100% coverage as a first pass. 

We could look at amending some of the logic in future, for example, extracting some of the code from __main__, which I have done as part of this PR in `visualise_results.py` to enable better testing. There are some methods that still have no coverage, for example the google_cloud_storage get_files method, as although it is possible to mock the GCS client is difficult to add meaningful tests.

Adds the tests to our actions workflow on PR and enforces coverage at 80%.

Adds isort to benchmark and runs as part of `make format`.

### How to review

Run the unit tests with `make test`. Check the tests added to make sure they make sense, can any other tests be added?

Check the benchmark is still working.


berroar:
[Add tests for eq-survey-runner-benchmark scripts (3 SP)](https://trello.com/c/WQxCvEQ0/5948-add-tests-for-eq-survey-runner-benchmark-scripts-3-sp)

scripts/visualise_results.py
```diff
@@ -8,35 +8,36 @@
...
+    try:
+        plt.style.use('fast')
 
-    plt.savefig('performance_graph.png', bbox_inches="tight")
-    print("Graph saved as performance_graph.png")
+        if (
+            number_of_days_to_plot and number_of_days_to_plot <= 45
+        ):  # To make the chart still easily digestible
+            df.plot.line(marker="o", markersize=8)
+            # plt.grid(True, axis="both", alpha=0.3)
```
rmccar, line 23: Are we keeping this line even though its commented out?

rmccar (commented):
Other than this looks good

scripts/visualise_results.py
rmccar: Yeah thought it must've been something like that

rmccar (approved):

VirajP1002 (approved):

Pipfile
```diff
@@ -17,6 +18,10 @@ google-cloud-storage = "*"
 pandas = "*"
 matplotlib = "*"
 slackclient = "*"
+pytest = "*"
```
liamtoozer, line 21: Does the pytest dependencies need to go into the dev-packages?

tests/test_visualise_results.py
```diff
@@ -0,0 +1,60 @@
...
+    mock_plot_data(dataframe, 1)
+
+    assert mock_plot_data.call_count == 1
+    assert_frame_equal(mock_plot_data.call_args[0][0], expected_data_frame)
+
+
+def test_plot_data(get_results_single_file):
+    dataframe = get_data_frame(get_results_single_file)
+    try:
+        plot_data(dataframe, 1)
```
liamtoozer, line 52: Quick couple of questions - Does this test need an assertion? Also, is it worth removing the file that gets generated here, or mocking the call and asserting that it's called?

scripts/visualise_results.py
```diff
@@ -8,35 +8,36 @@
...
 
-    plt.savefig('performance_graph.png', bbox_inches="tight")
-    print("Graph saved as performance_graph.png")
+        if (
+            number_of_days_to_plot and number_of_days_to_plot <= 45
+        ):  # To make the chart still easily digestible
+            df.plot.line(marker="o", markersize=8)
+            plt.grid(True, axis="both", alpha=0.3)
+        else:
+            df.plot.line()
```
liamtoozer, line 25: Another quick question! I noticed that this condition is missing test coverage, is this one of the bits that can't be tested?

Pipfile
berroar: Moved these now 👍

tests/test_visualise_results.py
berroar: I had a look and it seems like one of the only ways to test the graph generation is by using a library that is able to compare the graph generated in the test against a stored image. But spoke with Mebin and we agreed it was enough to test that the plot_data method does not fail when it is passed a valid dataframe for now. However as you pointed it was generating the graph when running the test, so I have mocked that and added an assertion that the `savefig` method is called 👍

scripts/visualise_results.py
berroar: This is one that needs to be looked at separately, i think we will raise a card to look at refactoring some of the benchmark scripts now that we have test coverage. Not sure that we ever use this else anymore.

tests/test_visualise_results.py
liamtoozer: Ah cool that makes sense, thanks!

scripts/visualise_results.py
liamtoozer: That makes sense, thanks 👍

liamtoozer (approved):
Tested locally and works great for me 👍

Pipfile
```diff
@@ -17,6 +21,8 @@ google-cloud-storage = "*"
 pandas = "*"
 matplotlib = "*"
 slackclient = "*"
+freezegun = "*"
+numpy = "*"
```
MebinAbraham, line 25: These should be dev packages.

berroar: Moved freezegun, removed numpy as I think I added it in error in the first place, it is already installed with pandas and not needed for the tests 👍

PR #153, opened 2024-04-24
Add script to post results to GitHub

petechd:
This adds new CI script to allow us to output benchmark results to Github PR comments. It is required for the wider Dependabot pipeline changes.

### How to review
Since it is integrated into the pipeline changes it needs to be tested as part of the pipeline.


scripts/benchmark_stats.py
```diff
@@ -8,8 +9,9 @@ class BenchmarkStats:
     PERCENTILES_TO_REPORT = (50, 90, 95, 99, 99.9)
     PERCENTILE_TO_USE_FOR_AVERAGES = 99
 
-    def __init__(self, folder_paths: List[str]):
+    def __init__(self, folder_paths: List[str], output_to_github: bool = False):
         self._files: List = []
+        self.output_to_github = output_to_github
```
berroar, line 14: Minor, I think if we use here we won't need to pass this new var through the classes 
```suggestion
        self.output_to_github = os.getenv("OUTPUT_TO_GITHUB", False)
```

berroar (commented):
Needs a PR description 😅

scripts/benchmark_stats.py
petechd: Added now and also removed the Github var from the slack ci script. Thanks.

berroar (approved):

liamtoozer (approved):
Seen this working in PRs, looks good to me! 👍

PR #154, opened 2024-04-25
Update payload to v2 layout to run daily benchmark

liamtoozer:
The daily benchmark needs to be updated to use the `v2` schema launch payload format to allow a successful run, following a recent Runner change to remove all traces of the `v1` schema launch.

The payload has now been altered to align with the example `v2` schema found in the [ons-schema-definitions](https://github.com/ONSdigital/ons-schema-definitions/blob/main/examples/rm_to_eq_runner/payload_v2/launch_jwt_business.json) repo (excluding the `sds_dataset_id`)

### How to review
Run the below benchmarks from this repo root, they should now run successfully. A follow-up ticket will be raised to investigate possibly removing the other JSON requests, as some appear to no longer be in use:

- `pipenv run ./run.sh requests/test_benchmark_business_happy_path.json`
- `pipenv run ./run.sh requests/test_benchmark_business_unhappy_path.json`
- `pipenv run ./run.sh requests/test_checkbox.json`

[Jira](https://jira.ons.gov.uk/browse/ECI-1254)

### DIFF
--- a/runner_benchmark/token_generator.py
+++ b/runner_benchmark/token_generator.py
@@ -14,25 +14,27 @@ SR_USER_AUTHENTICATION_PUBLIC_KEY_KID = 'e19091072f920cbf3ca9f436ceba309e7d814a6
 KEYS_FOLDER = './jwt-test-keys'
 
 PAYLOAD = {
-    'user_id': 'benchmark-user',
-    'period_str': 'July 2019',
-    'period_id': '201907',
-    'collection_exercise_sid': str(uuid4()),
-    'case_id': str(uuid4()),
-    'case_type': 'HI',
-    'display_address': '68 Abingdon Road, Goathill',
-    'ru_ref': '123456789012A',
-    'ru_name': 'Integration Testing',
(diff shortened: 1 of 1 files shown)


runner_benchmark/token_generator.py
```diff
@@ -81,7 +83,7 @@ def _get_payload_with_params(schema_name, schema_url=None, **extra_payload):
         datetime.now(tz=timezone.utc) + timedelta(days=7)
     ).isoformat()  # 7 days from now in ISO 8601 format
     for key, value in extra_payload.items():
-        payload_vars[key] = value
+        payload_vars['survey_metadata']['data'][key] = value
```
MebinAbraham, line 86: This wouldn't currently work. Our current use for this is for top level data (schema_url).
You would need one for extra top level data and extra survey metadata.

liamtoozer: Ah okay, will update now - thanks

berroar (approved):
Tested benchmark 👍

PR #158, opened 2024-06-24
Spike: Enable minify HTML for Runner

liamtoozer:
Run a performance investigation to assess whether we can re-enable the minification of HTML in Runner.

### How to review
* Read through the `summary.md` file - Is the investigation okay and does it answer our questions about enabling HTML minify?
* Should we run any additional benchmarks? Have I missed anything?

[Jira](https://jira.ons.gov.uk/browse/ECI-1162)


berroar (commented):
As discussed I have had a read through and looks like the findings here seem to align with the findings we did on the initial spike when we investigated enabling minify. As there are no visible benefits and some minor negative impacts I cannot at the moment see a reason to enable minify as we should expect to see performance improve when it is enabled. Will prob need to document in the decision log (providing other devs agree).

berroar (approved):

VirajP1002 (approved):
Looks good to me! 👍 I agree with the decision to not enable minify as Rhys mentioned above

petechd (approved):

PR #162, opened 2024-07-10
Migrate to docker image repo

liamtoozer:
We're migrating eQ pipeline resources to point at our centralised custom Docker file repo: https://github.com/ONSdigital/eq-runner-docker-images

### How to review
Push up each of the Dockerfiles from above, and set your dev pipeline `image_registry` var to point to your dockerfile registry. The pipeline job should still pass.


ci/run-benchmark.yaml
```diff
@@ -2,7 +2,7 @@ platform: linux
 image_resource:
   type: registry-image
   source:
-    repository: ((image_registry))/eq-app-k8s-deploy-image
```
petechd, line 5: This isn't directly mentioned on the card, only `eq-app-deploy-image` so I need to double check.

liamtoozer: Yeah, it's a bit odd as I couldn't spot which of our repos were creating the image named `eq-app-k8s-deploy-image` 🤔

VirajP1002 (approved):

PR #164, opened 2024-07-29
Add session/download-pdf endpoints and add graph to represent values

VirajP1002:
To update the benchmark outputs in order for us to monitor and highlight the performance of more intensive endpoints by including `session` and `download-pdf`. 

`benchmark_stats` has been provisioned with the 99th percentiles for the `session` and `download-pdf` endpoints. In order to achieve this, the happy path has been updated to include `download-pdf` such that the csv file created when running the benchmark contains the percentiles for this endpoint. The output "string" of the benchmark stats has been updated to include the new metrics for both Github and non-Github outputs.

`visualise_results` has been updated to support the new metrics. Currently, as the 99th percentiles for both `session` and `download-pdf` are significantly larger than the original figures, it was decided that outputting 2 graphs was the most appropriate solution. Thus, 2 dataframes have been created, one for the original percentiles and a new one for the additional metrics. The "image" of the graphs created use subplots to place both graphs created next to each other.

Various tests have been updated and parametrised. In addition to this `conftest` was updated to include more fixtures that support the new endpoints.

### How to review
- Check if the changes make sense
- Suggestions for refactoring
- Ensure all tests pass
- Use the [README](https://github.com/ONSdigital/eq-survey-runner-benchmark?tab=readme-ov-file#visualising-benchmark-results) to test the updated `visualise_results` script
- Spin up a benchmark pipeline and test the output to Slack and spin up a Dependabot staging pipeline to test the output to Github


scripts/benchmark_stats.py
```diff
@@ -89,6 +102,16 @@ def _process_file_data(self):
                         )
                         self._requests[row["Type"]]["total"] += request_count
 
+    @property
+    def pdf_percentile(self) -> Any:
+        if self._pdf_percentile:
+            return int(sum(self._pdf_percentile) / len(self._pdf_percentile))
```
petechd, line 108: `int(x)` only crops decimals so better if we use `round(x)` or if you want to math.floor(x) or math.ceil(x), but round(x) does both depending on the situation.

scripts/benchmark_stats.py
```diff
@@ -89,6 +102,16 @@ def _process_file_data(self):
...
+        return None
+
+    @property
+    def session_percentile(self) -> int:
+        return int(sum(self._session_percentile) / len(self._session_percentile))
```
petechd, line 113: Same as [the other comment](https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/164/files#r1714766385).

scripts/benchmark_stats.py
```diff
@@ -69,6 +76,12 @@ def _process_file_data(self):
                         )
                         self._total_failures += int(failure_count)
                     else:
+                        if row["Name"] == "/submitted/download-pdf":
+                            self._pdf_percentile.append(int(row.get("99%")))
```
petechd, line 80: Same as [the other comment](https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/164/files#r1714766385). Wouldn't it be better to use `round(x)` since it's more accurate than just cropping? Also Line 77 should be fixed as well?

scripts/benchmark_stats.py
```diff
@@ -69,6 +76,12 @@ def _process_file_data(self):
...
+
+                        if row["Name"] == "/session":
+                            self._session_percentile.append(int(row.get("99%")))
```
petechd, line 83: Same as [the other comment](https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/164/files#r1714766385).

scripts/benchmark_stats.py
```diff
@@ -89,6 +102,16 @@ def _process_file_data(self):
                         )
                         self._requests[row["Type"]]["total"] += request_count
 
+    @property
+    def pdf_percentile(self) -> Any:
```
petechd, line 106: Can this be made more accurate so we don't return `Any`, for example you can have default return here as well that returns `None` so we have better type?

scripts/benchmark_stats.py
```diff
@@ -39,6 +41,8 @@ def __str__(self):
                 f'{formatted_percentiles}<br />'
                 f'GETs (99th): {self.average_get}ms<br />'
                 f'POSTs (99th): {self.average_post}ms<br /><br />'
+                f'PDF: {str(self.pdf_percentile)+"ms" if self.pdf_percentile else "N/A"}<br />'
```
petechd, line 44: Simpler way to do this would be to make pdf_percentile return string with `ms` already attached since you do this conversion everywhere anyway (line 44, 58)?  so then it would be as simple as: `PDF: {self.pdf_percentile or "N/A"}<br />` so it choose "N/A" when `self.pdf_percentile` has falsey value.

scripts/visualise_results.py
```diff
@@ -7,37 +7,59 @@
...
+        plot_data(dataframes, number_of_days_to_plot)
+        plt.savefig(filename, bbox_inches="tight")
+        print("Graph saved as", filename)
     except Exception as e:
         raise GraphGenerationFailed from e
 
 
-def get_data_frame(results):
+def create_dataframe(result_fields, values_to_plot, prefix=True):
+    if prefix:
```
berroar, line 51: Is this referring to the suffix rather than a prefix? I think we can get rid of this if and the passing of the var if we create a percentiles_to_plot array with the `th` included (or something similar)

petechd:
You could possibly add some more context to the PR title, something like "... and graph splitting", "... performance graph changes", "... and results visualisation changes" or  "... to outputs", just something to give us more context when we check the squashed commits after.

berroar:
Got some odd behaviour with the output graph when the `Number_of_days` param is not set, may need to have a look e.g.
https://files.slack.com/files-pri/T04RP3ZJ3QF-F07GLLK7TNH/screenshot_2024-08-13_at_11.07.38.png

petechd:
As we discussed, maybe add a test where we can see "N/A" is outputted when `self. pdf_percentile` is  empty?

VirajP1002:
> Got some odd behaviour with the output graph when the `Number_of_days` param is not set, may need to have a look e.g. https://files.slack.com/files-pri/T04RP3ZJ3QF-F07GLLK7TNH/screenshot_2024-08-13_at_11.07.38.png

Addressed this, should be fixed now 👍

scripts/benchmark_stats.py
VirajP1002: Updated 👍

scripts/benchmark_stats.py
VirajP1002: Updated 👍

scripts/benchmark_stats.py
VirajP1002: Not sure for line 77 as it should be a whole number? Updated line 80 however 👍

scripts/benchmark_stats.py
VirajP1002: Updated 👍

scripts/benchmark_stats.py
VirajP1002: Updated 👍

scripts/benchmark_stats.py
VirajP1002: Updated to use a new function to format the string 👍

scripts/visualise_results.py
VirajP1002: Updated 👍

VirajP1002:
> As we discussed, maybe add a test where we can see "N/A" is outputted when `self. pdf_percentile` is empty?

There's a test to address this, when `self.pdf_percentile` is empty, it will return None rather than "N/A" and then another test is used to check if "N/A" is part of the response, not sure if that's sufficient?

scripts/benchmark_stats.py
petechd: Did you push the changes you've showed me? I can still see the `Any` type hint in the file.

petechd:
> You could possibly add some more context to the PR title, something like "... and graph splitting", "... performance graph changes", "... and results visualisation changes" or "... to outputs", you will come up with something better. Just something to give us more context when we check the squashed commits after.

Title looking great now!

scripts/benchmark_stats.py
berroar: This is useful but can it be extended to just be format_percentiles so it handles the adding of `ms` for all of them? Would make it more consistent then

VirajP1002: Took a slightly different approach that just updating the method, happy to change to an alternative if you had another solution in mind 👍

scripts/benchmark_stats.py
VirajP1002: Pushed this up now 👍

scripts/benchmark_stats.py
```diff
@@ -48,8 +58,11 @@ def __str__(self):
             f"Percentile Averages:\n"
             f"{formatted_percentiles}\n"
             f"---\n"
-            f"GETs (99th): {self.average_get}ms\n"
-            f"POSTs (99th): {self.average_post}ms\n"
+            f"GETs (99th): {formatted_metrics.get('average_get')}\n"
```
berroar, line 61: Can get rid of the .gets here as well 👍

scripts/visualise_results.py
```diff
@@ -6,38 +6,56 @@
...
 
-def plot_data(df, number_of_days_to_plot):
-    try:
-        plt.style.use("fast")
+def plot_data(dataframes, number_of_days_to_plot):
+    plt.style.use("fast")
+    fig, axes = plt.subplots(nrows=1, ncols=2, figsize=(15, 6))
 
+    for i in range(len(dataframes)):
+        plt.subplot(1, 2, i + 1)
```
berroar, line 23: Might need a comment to make it clear what this line is doing

petechd:
Do we want to fix the other places where we cast integer on floats and round up/down instead (as part of this PR, since we are already here)? Are there other places that still do that? Do other devs have any thoughts?

VirajP1002:
> Do we want to fix the other places where we cast integer on floats and round up/down instead (as part of this PR, since we are already here)? Are there other places that still do that? Do other devs have any thoughts?

I'm happy to add those in as it removes the rounding issue we talked about. These changes are minor and I guess it would be potentially tweaking the graph values? So I think it's somewhat relevant to this PR, but yep, let's see what the other devs think

scripts/benchmark_stats.py
VirajP1002: Removed 👍 there are a few others in that file from before, would it be worth removing those? e.g. `row.get("request_count")` or is a chance that it may not exist?

scripts/visualise_results.py
VirajP1002: Added that in 👍

scripts/benchmark_stats.py
berroar: Depends if we know that they will always be present

tests/test_visualise_results.py
```diff
@@ -3,60 +3,160 @@
...
-        plot_data(dataframe, 1)
+        create_graph([dataframe], 1, image_name)
         assert graph_output.call_count == 1
     except GraphGenerationFailed:
         pytest.fail("Graph generation failed")
 
 
-def test_plot_data_failed():
+@pytest.mark.parametrize(
+    "image_name",
```
berroar, line 136: Is there any point parametrising these? They are just strings

scripts/visualise_results.py
```diff
@@ -49,16 +70,33 @@ def get_data_frame(results):
...
-    results = get_results(folders, number_of_days)
-    dataframe = get_data_frame(results)
 
-    plot_data(dataframe, number_of_days)
+    results = list(get_results(folders, number_of_days))
+
+    performance_dataframe = get_performance_data_frame(results)
+    additional_metrics_dataframe = get_additional_metrics_data_frame(results)
+
+    dataframe_values = [performance_dataframe, additional_metrics_dataframe]
```
berroar, line 100: Should this just be `dataframes`? Do we need this var since it's only used once, maybe would be clearer if the method call used the argument name?

scripts/benchmark_stats.py
```diff
@@ -30,15 +32,25 @@ def __str__(self):
             f"{percentile}th: {self.percentiles[percentile]}ms"
             for percentile in self.PERCENTILES_TO_REPORT
         )
+        formatted_metrics = {
```
berroar, line 35: Could the formatted metrics and percentiles maybe move to the init after the file is processed? Might make this a bit tidier 🤔

tests/test_visualise_results.py
```diff
@@ -3,60 +3,160 @@
...
+    ),
+)
+def test_individual_graph_creation(
+    results_file, data_frame_method, image_name, mocker, request
+):
+    dataframe = data_frame_method(request.getfixturevalue(results_file))
     try:
         graph_output = mocker.patch("matplotlib.pyplot.savefig")
-        plot_data(dataframe, 1)
+        create_graph([dataframe], 1, image_name)
```
berroar, line 129: The behaviour is a bit weird now that even if you only attempt to plot 1 graph, the output image will still have 2 🤔

tests/test_visualise_results.py
petechd: I guess this makes sense so you don't need to create the for loop in the test?

VirajP1002: That was the reason why I did it, like Pete said, however, I do understand that there's no point in have the image name parametrised since the filename could be anything, so we could have loads of filenames and it won't make a difference to the test

scripts/visualise_results.py
VirajP1002: I kept the var in so it makes the method call a bit clearer, happy to just put the array in the method call, I updated the vars to keep the names consistent 👍

tests/test_visualise_results.py
VirajP1002: The current logic creates two graphs because of the subplots. We'll have to keep subplots in order to have 2 graphs. Currently, for this test, we only intend to create "one graph", but there will be 2 subplots, the other subplot isn't populated with the data simply because there's no dataframe for the second subplot, so I guess this test may need to be changed or deleted because there cannot be only one graph where 2 subplots are used.

berroar: Yeah similarly if you pass more than 2 dataframes to the method it breaks at the moment. This test is hinting that we are can handle a single dataframe when the method has been designed to render the same two dataframes each time.

tests/test_visualise_results.py
berroar: @petechd But the actual value doesn't matter to the test in this instance, it is just a string and could be anything?

tests/test_visualise_results.py
berroar: I think the code either needs to be clear that this method plots the two expected dataframes, or it needs to be made generic so if we pass 1 or multiple dataframes, it plots whatever is there.

scripts/benchmark_stats.py
VirajP1002: Updated 👍

tests/test_visualise_results.py
VirajP1002: Updated 👍

berroar:
We should bump the test coverage as part of this PR as I think we are up to 85% now 🎉

berroar (approved):

PR #165, opened 2024-07-30
Update to use poetry package manager

petechd:
### What is the context of this PR?
We migrate our remaining repositories from Pipenv to Poetry for consistency in our eQ ecosystem. Updating libraries resulted in some changes in code formatting that I had to include in this PR.

### How to review
All scripts/actions that are affected and the dockerised app needs to be thoroughly tested, check for any unnecessary additions/repetition and if nothing was omitted.

PR #167, opened 2024-08-01
Update libraries and use spawn rate

petechd:
This replaces Locust `hatch rate` option with `spawn rate` that was introduced in [this PR](https://github.com/locustio/locust/pull/1523). After the update to the latest version `hatch rate` is no longer recognised and benchmark only spawns one user. This is also needed for the [Poetry migration PR](https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/165).

### How to review
Run benchmark using `spawn rate` locust option, check if it finishes with no logs errors and successful submissions.


berroar (commented):
Also needs a pipelines update as the var is currently `users_hatch_rate: '50'`, so we can align

berroar (approved):

petechd:
> Also needs a pipelines update as the var is currently `users_hatch_rate: '50'`, so we can align

[Now added pipelines PR](https://github.com/ONSdigital/eq-pipelines/pull/616)

VirajP1002 (approved):

rmccar (approved):

PR #173, opened 2024-09-23
Bump python version & dependencies

liamtoozer:
We need to upgrade to Python 3.12 and bump our dependencies to ensure we are up-to-date with important features, security patches etc. In addition, extra bits added were:

* Added `--only main` to poetry install command in Dockerfile
* Move `RUN` grouping to single image layer

### How to review
Ensure the build passes and that you're able to run locally and with Docker

### Checklist
* [ ] eq-translations updated to support any new schema keys which need translation


Dockerfile
```diff
@@ -9,9 +9,9 @@ RUN apt-get update && apt-get install -y gcc python3-dev
 # Install the required dependencies via pip
 COPY pyproject.toml pyproject.toml
 COPY poetry.lock poetry.lock
-RUN pip install "poetry==1.8.3"
-RUN poetry config virtualenvs.create false
-RUN poetry install
+RUN pip install "poetry==1.8.3" \
+    && poetry config virtualenvs.create false \
+    && poetry install --only main
```
petechd, line 14: `--without dev` would be probably more self explanatory but we use `--only main` in other projects so we can keep it as it is.

petechd (approved):

Dockerfile
liamtoozer: I was getting a deprecation warning when I was testing this unfortunately 
```
The `--no-dev` option is deprecated, use the `--only main` notation instead.
```
I agree though, it reads better to me too 😢

berroar (approved):

Yuyuutsu (approved):

Dockerfile
liamtoozer: Ah sorry, I misread your original comment 🤦 I've tried with `--without dev` and it still works (without deprecation warning), so will change to that now, thanks

petechd (approved):

Dockerfile
liamtoozer: I think that `only main` just installs main dependencies, whereas `without dev` installs _everything_ except dev dependencies. 

So it might be safer to keep it as `only main` for now, just in case at some point we end up adding another dependency group?

PR #181, opened 2025-01-06
Switch to FastHttpUser

petechd:
Result of our investigation into latency issues. It was one of the recommendations in the original Locust docs that we decided to test. Link to the topic in the docs: **https://docs.locust.io/en/stable/increasing-request-rate.html#load-generation-performance**.

### How to review
Run a benchmark pipeline and check the performance graphs.


Yuyuutsu (approved):
Investigation looks good and Benchmark runs as expected.

berroar (approved):

PR #185, opened 2025-05-15
Create codeql-analysis.yml

Yuyuutsu:
We want to be consistent in all eQ repositories.
This PR is to change our codeQL settings from the default to advanced to align with Runner

### How to review
Check CodeQL YML settings make sense
CodeQL Scan runs as expected


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,43 @@
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '22 16 * * 5'
```
petechd, line 9: Specific time and day of the week, issue in multiple related PRs.

Yuyuutsu: I shall make the changes on all the PR's now 👍

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+      matrix:
+        include:
+        - language: actions
+          build-mode: none
+        - language: python
+          build-mode: none
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
rmccar, line 33:
```suggestion
      uses: actions/checkout@4.2.2
```
Couldn't we just set it like this here? Same with the others here and on multiple PRs

rmccar (approved):

PR #104, opened 2022-02-16
Document runner's PDF performance under Weasyprint and PDFKit

MebinAbraham:
### What is the context of this PR?
Documents the performance of runner under two PDF libraries, Weasyprint and PDFKit (current implementation).

### How to review 

- Does the document accurately reflect the test?
- Are the observations accurate?
- Can any other observations be made?
- Do you agree with the recommendations?
- Should any other recommendations be made?

PR #196, opened 2026-02-05
EQS-512: Spike - Alternatives to pdfkit performance investigation

liamtoozer:
The `pdfkit` package and `wkhtmltopdf` engine used in Runner to export HTML to PDF are now deprecated. 

There was a Spike [carried out about 3 years ago](https://github.com/ONSdigital/eq-survey-runner-benchmark/pull/104) to look into moving to `weasyprint`. At the time, it was found that `weasyprint` was not as efficient as exporting to PDF as our current `pdfkit`. However, things have since moved on for `weasyprint`, and the deprecation for `pdfkit` was put in place just 2 years ago, so may be worth another look.

Part of this spike also looks at the `xhtml2pdf` package as well

### How to review
* Read through the `doc/performance-investigations/0025-pdfkit-alternatives/summary.md`. As mentioned in the previous Spike, think about the following:
  * Does the document accurately reflect the test?
  * Are the observations accurate?
  * Can any other observations be made?
  * Do you agree with the recommendations?
  * Should any other recommendations be made?


farres1:
The write-up looks good and summarises the information clearly. If we can dedicate time to researching Option 4, I agree with your conclusion of Options 2, 4 and 7. Otherwise, I think combining Options 2 and 7 would be the most beneficial approach.

doc/performance-investigations/0025-pdfkit-alternatives/summary.md
```diff
@@ -0,0 +1,220 @@
...
+|-----------------------------------------|---------------------|------------------|-------------------|----------------------|----------------------|----------------------|----------------------|------------------------|-----------------------|-----------------------------------|----------------|----------------|------------------------------------------------------|
+| Baseline with pdfkit                    | 57.60               | 362              | 133               | 1600                 | 1900                 | 2500                 | 3600                 | 4100                   | 4100                  | 30                                | 34,445         | 0              | [outputs](outputs/baseline/2026-01-26T13:41:05)      |  
+| xhtml2pdf with custom CSS file (Run 1)  | 58.60               | 305              | 126               | 500                  | 640                  | 670                  | 970                  | 990                    | 990                   | 22                                | 35,032         | 0              | [outputs](outputs/investigation/2026-01-28T16:12:30) |
```
farres1:
```suggestion
| xhtml2pdf with custom CSS file (Run 1)  | 58.60               | 305              | 126               | 500                  | 640                  | 760                  | 970                  | 990                    | 990                   | 22                                | 35,032         | 0              | [outputs](outputs/investigation/2026-01-28T16:12:30) |
```

From the document, 95th percentile is 760.

liamtoozer: Good spot 👍  Updated, thanks

Yuyuutsu:
Really nice investigation!
I think the next step we choose is really depends on the resources we have. I would like us to look into our own PDF generation but that is a monumental task. Even changing to weasyprint, which looks like our best option, is still a really big task. I do like the idea of incorporating a loop icon for better UX.

This is most likely something we may need to discuss as a whole team.

farres1 (approved):

rmccar (approved):
Looks good. One thought is that for option 7 on line 208 you mention needing DS and accessibility considerations for adding the wait loop icon, we actually already have a button in the DS that displays a wait loop icon when clicked so this might not be such an issue

PR #197, opened 2026-02-11
EQS-612 - Update Protobuf

farres1:
Upgrades Protobuf version.

https://officefornationalstatistics.atlassian.net/browse/EQS-612

### How to review

**Check:**
- `poetry.lock` uses `Protobuf` version `6.33.5`


pyproject.toml
```diff
@@ -16,6 +16,7 @@ google-cloud-storage = "*"
 pandas = "*"
 matplotlib = "*"
 slack_sdk = "*"
+protobuf = "6.33.5"
```
VirajP1002: Just had a quick look here, and I don't we normally pin dependencies here, so I believe `protobuf` shouldn't be pinned? Otherwise it won't update or is there another reason why?

Also, is this a top-level dependency (not a dependency of another)? If so, this shouldn't be added to the `toml` file? I did have a look and I believe it is top-level, but just wanted to double check 😅

liamtoozer (approved):

Farhanam76 (approved):

